PR #5, opened 2021-11-17
feat: add price-oracle traits

Roznovjak:


traits/src/lib.rs
```diff
@@ -165,3 +166,33 @@ pub trait ShareTokenRegistry<AssetId, AssetName, Balance, Error>: Registry<Asset
...
+	fn on_trade(amm_transfer: &AMMTransfer<AccountId, AssetId, AssetPair, Balance>, liq_amount: Balance);
+	/// Known overhead for a trade in `on_initialize/on_finalize`.
+	/// Needs to be specified here if we don't want to make AMM pools tightly coupled with the price oracle pallet, otherwise we can't access the weight.
+	/// Add this weight to an extrinsic from which you call `on_trade`.
+	fn on_trade_weight() -> Weight;
+}
+
+impl<AccountId, AssetId, AssetPair, Balance> OnTradeHandler<AccountId, AssetId, AssetPair, Balance> for () {
+	fn on_trade(_amm_transfer: &AMMTransfer<AccountId, AssetId, AssetPair, Balance>, _liq_amount: Balance) {}
+	fn on_trade_weight() -> Weight {
```
enthusiastmartin, line 194: what is the purpose of this method ?

Roznovjak: it is stated above: 
```
Known overhead for a trade in `on_initialize/on_finalize`.
Needs to be specified here if we don't want to make AMM pools tightly coupled with the price oracle pallet, otherwise we can't access the weight.
Add this weight to an extrinsic from which you call `on_trade`.
```

enthusiastmartin: ehm. I am not sure if i understand its purpose.  How it used in xyk ?

Roznovjak: [That's](https://github.com/galacticcouncil/Basilisk-node/blob/17f54ed443a54adc05eab44b740507db2995b253/pallets/price-oracle/src/lib.rs#L259) how we set the weight.
You can see [here](https://github.com/galacticcouncil/Basilisk-node/blob/17f54ed443a54adc05eab44b740507db2995b253/pallets/xyk/src/lib.rs#L528) how it is used in the xyk pallet

enthusiastmartin (approved):

PR #9, opened 2021-11-22
feat: get share token from amm

martinfridrich:
Liquidity mining pallet need to know share token from amm `AssetPair`.


enthusiastmartin (changes requested):
The pallet version has to be increased.

lumir-mrkva (approved):

PR #10, opened 2021-11-25
feat: locked balance trait

lumir-mrkva:


traits/src/lib.rs
```diff
@@ -198,3 +199,7 @@ impl<AccountId, AssetId, AssetPair, Balance> OnTradeHandler<AccountId, AssetId,
 pub trait CanCreatePool<AssetId> {
     fn can_create(asset_a: AssetId, asset_b: AssetId) -> bool;
 }
+
+pub trait LockedBalance<AssetId, AccountId, Balance> {
+    fn get_by_lock(asset: AssetId, account: AccountId, lock: LockIdentifier) -> Balance;
```
enthusiastmartin, line 204: I would follow similar param order and naming as it is in the lockable traits for set, extend and remove lock

```
fn set_lock(lock_id: LockIdentifier, currency_id: AssetId, who: AccountID, amount: Balance) {.. 
```

That means:

```suggestion
    fn get_by_lock(lock_id: LockIdentifier, currency_id: AssetId, who: AccountId) -> Balance;
```

PR #11, opened 2021-12-03
update types in price-oracle traits

Roznovjak:


lumir-mrkva (changes requested):
bump versions

traits/src/lib.rs
```diff
@@ -171,29 +171,29 @@ pub trait AssetPairAccountIdFor<AssetId, AccountId> {
...
+impl<AssetId> OnCreatePoolHandler<AssetId> for () {
+    fn on_create_pool(_asset_a: AssetId, _asset_b: AssetId) {}
 }
 
 /// Handler used by AMM pools to perform some tasks when a trade is executed.
-pub trait OnTradeHandler<AccountId, AssetId, AssetPair, Balance> {
+pub trait OnTradeHandler<AssetId, Balance> {
     /// Include a trade in the average price calculation of the price-oracle pallet.
-    fn on_trade(amm_transfer: &AMMTransfer<AccountId, AssetId, AssetPair, Balance>, liq_amount: Balance);
+    fn on_trade(asset_a: AssetId, asset_b: AssetId, amount_in: Balance, amount_out: Balance, liq_amount: Balance);
```
lumir-mrkva, line 188: why was this changed? it is possible that you'll need fee information in order to calculate price inn some cases

lumir-mrkva: but you can get around it by supplying amended amout_in / out.

Roznovjak: because when we added the price-oracle pallet to the Basilisk repo, the compiler treated `AssetId` and `AccountId` types as distinct from the ones used in the Basilisk repo.

PR #13, opened 2021-12-15
feat: get fee from amm trait

Roznovjak:
Required by https://github.com/galacticcouncil/Basilisk-node/issues/262

### DIFF
--- a/traits/src/lib.rs
+++ b/traits/src/lib.rs
@@ -43,7 +43,7 @@ pub struct AMMTransfer<AccountId, AssetId, AssetPair, Balance> {
 }
 
 /// Traits for handling AMM Pool trades.
-pub trait AMM<AccountId, AssetId, AssetPair, Amount: Zero, Fee> {
+pub trait AMM<AccountId, AssetId, AssetPair, Amount: Zero> {
     /// Check if both assets exist in a pool.
     fn exists(assets: AssetPair) -> bool;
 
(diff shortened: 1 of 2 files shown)


traits/src/lib.rs
```diff
@@ -117,6 +117,8 @@ pub trait AMM<AccountId, AssetId, AssetPair, Amount: Zero> {
     fn get_max_in_ratio() -> u128;
 
     fn get_max_out_ratio() -> u128;
+
+    fn get_fee(_maybe_pool_account_id: Option<&AccountId>) -> Option<Fee>;
```
enthusiastmartin, line 121: Does it need to be all optional ? 

I would just make the parameter pool id not optional. And result too - if pool does not exist, return 0. 

You alway want to get a fee for a pool. I cant think of a scenario where you would need to call get fee with None. But i may be wrong ?

Roznovjak: In the case of XYK pools, every pool has the same fee, and the pool id parameter is not necessary. But you are right, it is probably less confusing without `Option`. But I would like to use `Option` as the return value because zero fee is valid fee value.

enthusiastmartin: sure but if you dont have pool ( pool does not exist ) , it does not matter if fee is zero or not.

traits/src/lib.rs
```diff
@@ -117,6 +117,8 @@ pub trait AMM<AccountId, AssetId, AssetPair, Amount: Zero> {
     fn get_max_in_ratio() -> u128;
 
     fn get_max_out_ratio() -> u128;
+
+    fn get_fee(pool_account_id: &AccountId) -> (u32, u32);
```
enthusiastmartin, line 121: I was thinking to make this as separate trait. 

Where is this used precisely ?

Roznovjak: It is only used in the exchange pallet

enthusiastmartin (approved):
Let's merge it for now. I am planning to refactor the AMM trait in near future anyway, so i might separate it if i see it necessary.

Issue #14, opened 2021-12-17
Remove dependecies on basilisk


green-jay:
- [ ] check for any basilisk/hydra dependencies and remove them

PR #16, opened 2021-12-27
build: remove dependencies

Roznovjak:
remove basilisk dependencies
Fixes: #14

### DIFF
--- a/price-oracle/Cargo.toml
+++ b/price-oracle/Cargo.toml
@@ -21,12 +21,9 @@ version = '2.3.1'
 serde = {features = ['derive'], optional = true, version = '1.0.101'}
 scale-info = { version = "1.0", default-features = false, features = ["derive"] }
 
-# TODO remove this dependency once the AssetPair struct is moved to this repository
-# Basilisk dependencies
-primitives = { git = "https://github.com/galacticcouncil/basilisk-node", rev = "c0bf049f0ee91c392f163768ef707c94c7981c46", default-features = false}
-
 # HydraDX dependencies
 hydradx-traits = { path = "../traits", default-features = false }
+hydra-dx-math = { git = "https://github.com/galacticcouncil/hydradx-math", rev = "ed5182a62741ab6d49b11c85290ddd05413cb666", default-features = false }
 
 # Substrate dependencies
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false, optional = true }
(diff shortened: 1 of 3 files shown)


price-oracle/Cargo.toml
```diff
@@ -21,10 +21,6 @@ version = '2.3.1'
 serde = {features = ['derive'], optional = true, version = '1.0.101'}
 scale-info = { version = "1.0", default-features = false, features = ["derive"] }
 
-# TODO remove this dependency once the AssetPair struct is moved to this repository
-# Basilisk dependencies
-primitives = { git = "https://github.com/galacticcouncil/basilisk-node", rev = "c0bf049f0ee91c392f163768ef707c94c7981c46", default-features = false}
-
 # HydraDX dependencies
 hydradx-traits = { path = "../traits", default-features = false }
```
green-jay, line 25: warehouse-traits/gcrl-traits? i think we should get rid of innacurate naming and keep this repo super agnostic

enthusiastmartin: I would go with `hydradx-traits`, so it is clear that it belongs to `hydradx `projects. It has nothing really to do with the warehouse. you use one or more crates from warehouse not the warehouse itself. 

the thing that you see this in code where you use those traits. eg.

`use hydradx_traits::*`;

price-oracle/Cargo.toml
```diff
@@ -34,15 +30,12 @@ frame-support = { git = "https://github.com/paritytech/substrate", branch = "pol
...
 frame-system-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false, optional = true }
 sp-core = { git = 'https://github.com/paritytech/substrate', branch = "polkadot-v0.9.12", default-features = false }
-sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false }
 sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false }
 
 [dev-dependencies]
 sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false }
-pallet-xyk = { git = "https://github.com/galacticcouncil/basilisk-node", rev = "c0bf049f0ee91c392f163768ef707c94c7981c46", default-features = false }
 orml-tokens = { default-features = false, git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "ee37b76a7a8cf159e6c41851942758746318bb84" }
 orml-traits = { default-features = false, git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "ee37b76a7a8cf159e6c41851942758746318bb84" }
```
enthusiastmartin, line 38: I believe you can remove the orml depedency too. It does not appear to be used anywhere ?

price-oracle/Cargo.toml
green-jay: AFAIU these traits will be used in both hydradx and basilisk = shared traits = belong to both projects. If it does not have anything to do with the warehouse why does it reside here?

enthusiastmartin: well. Here ( within this repo)  i consider basilisk and hydradx both as hydradx projects.  

i would prefer to have the trait crate easily identifiable , warehouse is just too generic ( eg. orml-traits )

If you happen to use a pallet from warehouse in another project, and you'd need traits too. it would be clear where it comes from when you do `use hydradx_traits`;

green-jay: Imho that is not semantically correct, basilisk and hydradx are two separate projects and belong to galactic council, not hydradx. Where it originates from is not important. Naming it hydradx is even misleading. That is why i suggested gcrl (grcl? :)-traits, or eg. galactic-traits. i would lean towards having shared things only here and avoid any dependency on basilisk or hydra

enthusiastmartin: Well, let's open an issue for discussion to rename hydradx-traits;-) 

It is not part of this PR anyway.

enthusiastmartin: #17  let's discuss there.

enthusiastmartin (approved):

green-jay (approved):

PR #19, opened 2022-01-10
feat:  transaction multi payment pallet

enthusiastmartin:
Added transaction multi payment pallet with support of spot price provider instead of direct need of amm pool.

Todo:
- [x] Load spot prices in on_init
- [ ] Update benchmarks


Roznovjak (approved):

transaction-multi-payment/src/lib.rs
```diff
@@ -0,0 +1,551 @@
...
+            let mut weight: Weight = 0;
+
+            for (asset_id, fallback_price) in <AcceptedCurrencies<T>>::iter() {
+                let maybe_price = T::SpotPriceProvider::spot_price(asset_id, native_asset);
+
+                let price = maybe_price.unwrap_or(fallback_price);
+
+                AcceptedCurrencyPrice::<T>::insert(asset_id, price);
+
+                weight += T::WeightInfo::get_spot_price();
```
Roznovjak, line 101: Where does this weight come from? I can't find the benchmarks.

traits/src/pools.rs
```diff
@@ -0,0 +1,11 @@
+pub trait SpotPriceProvider<AssetId> {
+    type Price;
+
+    fn pair_exists(asset_a: AssetId, asset_b: AssetId) -> bool;
```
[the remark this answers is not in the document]
enthusiastmartin: yes/no, it is possible to go without that but i have a feeling that it may come handy at some point. 

And It is already used in the XYK implementation of the provider.

transaction-multi-payment/src/lib.rs
enthusiastmartin: benchmarks must be written for given specific asset type - so it is not part of the pallet now as it is just generic ( so it not possible to set it up with multiple assets ). 

Benchmarks are in basilisk runtime.

Roznovjak (approved):

Issue #20, opened 2022-01-12
price-oracle: remove panic from on_finalize


enthusiastmartin:
There is a panic call in the on_finalize.

on_finalize cannot panic.

PR #21, opened 2022-01-13
refactor: remove panic from on_finalize

Roznovjak:
fixes: #20

### DIFF
--- a/price-oracle/src/lib.rs
+++ b/price-oracle/src/lib.rs
@@ -143,10 +143,7 @@ pub mod pallet {
                 *value = value
                     .saturating_add(
                         Self::new_assets()
-                            .len()
-                            .try_into()
-                            // we can use unwraps here because on_create_pool verifies that new_assets returns u32 and we can safely add this value to TrackedAssetsCount
-                            .unwrap(),
+                            .len() as u32
                     )
             });
 
--- a/price-oracle/src/lib.rs
+++ b/price-oracle/src/lib.rs
@@ -141,14 +141,13 @@ pub mod pallet {
             // add newly registered assets
             let _ = TrackedAssetsCount::<T>::mutate(|value| {
                 *value = value
-                    .checked_add(
+                    .saturating_add(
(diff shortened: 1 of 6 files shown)


price-oracle/src/lib.rs
```diff
@@ -133,20 +139,17 @@ pub mod pallet {
...
+                            // we can use unwraps here because on_create_pool verifies that new_assets returns u32 and we can safely add this value to TrackedAssetsCount
+                            .unwrap(),
                     )
-                    .ok_or(Error::<T>::PriceComputationError)?;
-                Ok(())
-                // We don't want to throw an error here because this method is used in different extrinsics.
-                // We also do not expect to have more than 2^32 assets registered.
-            })
-            .map_err(|_| panic!("Max number of assets reached!"));
+                    .unwrap()
```
enthusiastmartin, line 151: Well, this does not help much -  you just moved the panic inside the unwrap;-) 

on finalize cannot throw any error, or panic, or anything. otherwise the chain will stall. 

you already ignore the result of this expression anyway, so just dont do unwrap at all.

price-oracle/src/lib.rs
```diff
@@ -133,20 +139,17 @@ pub mod pallet {
             PriceDataAccumulator::<T>::remove_all(None);
 
             // add newly registered assets
-            let _ = TrackedAssetsCount::<T>::try_mutate(|value| -> Result<(), DispatchError> {
+            let _ = TrackedAssetsCount::<T>::mutate(|value| {
                 *value = value
                     .checked_add(
                         Self::new_assets()
                             .len()
                             .try_into()
```
enthusiastmartin, line 147: Like this ? 

```suggestion
                    .checked_add(
                        Self::new_assets()
                            .len() as u32
                           
```

price-oracle/src/lib.rs
enthusiastmartin: see suggestion above

PR #40, opened 2022-03-31
feat: add nft pallet

green-jay:


nft/src/weights.rs
```diff
@@ -0,0 +1,85 @@
+// This file is part of Basilisk-node.
```
enthusiastmartin, line 1: this should be part galactic warehouse

also other files are missing copyright section.

transaction-multi-payment/src/lib.rs
```diff
@@ -561,9 +561,9 @@ where
         match call.is_sub_type() {
             Some(Call::set_currency { currency }) => match Pallet::<T>::check_balance(who, *currency) {
                 Ok(_) => Ok(()),
-                Err(error) => Err(TransactionValidityError::Invalid(
-                    InvalidTransaction::Custom(error.as_u8()),
-                )),
+                Err(error) => Err(TransactionValidityError::Invalid(InvalidTransaction::Custom(
```
enthusiastmartin, line 564: wait? what ? why changing multi payment pallet here ? 

it is likely just reformat but still. i would revert it so it is not confusing.

enthusiastmartin (commented):
just two comments, other than that it is ok and can be merged.

PR #42, opened 2022-04-06
feat: transfer tx fees to treasury

Roznovjak:


enthusiastmartin:
i would suggest to keep the original implementation where the native is withdrawn and refunds are deposited.

and introduce new one with fees transferred to selected one.

in this way - it will be easily to swap back if needed.

enthusiastmartin:
@apopiak Could you have a look at this ?

the reason is that it seems that this PR introduces something similar to what you'd need for xcm trader ( see the TransactioPaymentDataProvider). 
So perhaps, we can/should consolidate this into one thing. 

Also, review of the PR by you would be appreciated too.

transaction-multi-payment/src/lib.rs
```diff
@@ -361,68 +400,120 @@ where
...
+                PaymentInfo::Native(paid_fee) => {
+                    let refund_amount = paid_fee.saturating_sub(corrected_fee);
+                    C::transfer(&fee_receiver, who, refund_amount, ExistenceRequirement::KeepAlive)
+                        .map_err(|_| TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
+                },
+                PaymentInfo::NonNative(paid_fee, currency, price) => {
+                    // calculate corrected_fee in the non-native currency
+                    let converted_corrected_fee = price.checked_mul_int(corrected_fee).ok_or(TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
+                    let converted_fee = paid_fee.saturating_sub(converted_corrected_fee);
+                    MC::transfer(currency.into(), &fee_receiver, who, converted_fee.into())
```
apopiak, line 504:
```suggestion
                    let refund = paid_fee.saturating_sub(converted_corrected_fee);
                    MC::transfer(currency.into(), &fee_receiver, who, refund.into())
```
nitpick: should be refund

transaction-multi-payment/src/lib.rs
```diff
@@ -361,68 +400,120 @@ where
...
+        already_withdrawn: Self::LiquidityInfo,
+    ) -> Result<(), TransactionValidityError> {
+        let fee_receiver = DP::get_fee_receiver().ok_or(TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
+
+        if let Some(paid) = already_withdrawn {
+            // Calculate how much refund we should return and transfer the amount to the account that paid the fees
+            match paid {
+                PaymentInfo::Native(paid_fee) => {
+                    let refund_amount = paid_fee.saturating_sub(corrected_fee);
+                    C::transfer(&fee_receiver, who, refund_amount, ExistenceRequirement::KeepAlive)
```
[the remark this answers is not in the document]
enthusiastmartin: How else would be possible to withdraw fee and avoiding such scenarios?  Would reserve some balance instead of transfer work ?

i mean, you'd first reserve and then only transfer the actual fee and unreserve the remaining balance ?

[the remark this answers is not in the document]
apopiak: The `MultiCurrency` trait [offers `withdraw` and `deposit`](https://github.com/open-web3-stack/open-runtime-module-library/blob/6a00fb20f0164b51764a06a43d0dccdaf6b758a2/traits/src/currency.rs#L67-L73) which could be used together similar to how Substrate's transaction payment is implremented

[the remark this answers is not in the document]
Roznovjak: The problem is that `Currency` and `MultiCurrency` traits are inconsistent. `Currency` returns `Imbalance` type, `MultiCurrency` returns `Balance` type, and it's impossible to combine them. We can't create an imbalance from some balance and we also can't get rid of imbalance without calling it's `drop` implementation. This is the main reason why we are using transfers in our implementation. We can handle each currency type separately and use `withdraw` methods, but then in case of any error during execution of an extrinsic, tokens are burned instead of deposited because `correct_and deposit_fee` is not called. And I'm not sure if we want this behavior.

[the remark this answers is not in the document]
Roznovjak: Is skipping the deposit in case `refund` is zero really needed? Both deposit methods (native and non-native) have a check for zero balance at the beginning.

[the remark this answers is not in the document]
apopiak: :shrug: good point, that's a fair assumption for the most common implementations.

transaction-multi-payment/src/lib.rs
```diff
@@ -182,34 +181,34 @@ pub mod pallet {
...
     #[pallet::getter(fn currency_price)]
     pub type AcceptedCurrencyPrice<T: Config> = StorageMap<_, Twox64Concat, AssetIdOf<T>, Price, OptionQuery>;
 
-    /// Account to use when pool does not exist.
+    /// Account where fees are deposited
     #[pallet::storage]
-    #[pallet::getter(fn fallback_account)]
-    pub type FallbackAccount<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
+    #[pallet::getter(fn fee_receiver)]
+    pub type FeeReceiver<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
```
jak-pan, line 206: Can we do this as constant? We'll save read on every tx. cc @lumir-mrkva

Roznovjak: fallback account has been removed from the storage and is now configurable in the Config

jak-pan (approved):

transaction-multi-payment/src/mock.rs
```diff
@@ -170,8 +173,24 @@ impl pallet_balances::Config for Test {
     type ReserveIdentifier = ();
 }
 
+pub struct DepositAll;
+
+impl DepositFee<AccountId, AssetId, Balance> for DepositAll {
```
lumir-mrkva, line 178: @Roznovjak can you provide this as default implementation, so it doesn't have to be implemented in the runtime?

Roznovjak: Would you like to have it implemented for the unit type `()` or just to have `DepositFee` struct publicly available from the pallet?

transaction-multi-payment/src/lib.rs
```diff
@@ -361,68 +386,161 @@ where
...
+            let mut fee_amounts = Vec::new();
+            if !fee.is_zero() {
+                fee_amounts.push((currency, fee));
+            }
+            if !tip.is_zero() {
+                fee_amounts.push((currency, tip));
+            }
+
+            DF::deposit_fee(&fee_receiver, fee_amounts.into_iter())
+                .map_err(|_| TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
```
enthusiastmartin, line 533: this goes together with the other comment about deposit_fee interface.

why this needs to be done in this way ? 

what about this ?
```suggestion
           let total  = fee + tip;
           if total > 0 {
                  DF::deposit_fee(&fee_receiver, currency, total)
                      .map_err(|_| TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
                }
```

transaction-multi-payment/src/lib.rs
```diff
@@ -361,68 +386,161 @@ where
...
+impl<T: Config> DepositFee<T::AccountId, AssetIdOf<T>, BalanceOf<T>> for DepositAll<T>
+where
+    AssetIdOf<T>: Ord,
+{
+    fn deposit_fee(who: &T::AccountId, amounts: impl Iterator<Item = (AssetIdOf<T>, BalanceOf<T>)>) -> DispatchResult {
+        // merge items with the same asset ID
+        let amounts = amounts
+            .sorted_unstable_by(|a, b| Ord::cmp(&a.0, &b.0))
+            .coalesce(|a, b| if a.0 == b.0 { Ok((a.0, a.1 + b.1)) } else { Err((a, b)) });
+
```
enthusiastmartin, line 421: what is the need to have the iterator ? 

my understanding is that  this is going to be used only to transfer fee + tip, which is always same currency.

```
fn deposit_fee(who, currency, amount) 
``` 

would not be sufficient ? 

See the other comment where deposit_fee is called.

transaction-multi-payment/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-transaction-multi-payment"
-version = "7.0.2"
+version = "7.1.0"
```
enthusiastmartin, line 3: i believe this should go version 8. as pallet config has changed and it is major breaking change?

transaction-multi-payment/src/lib.rs
```diff
@@ -181,35 +185,26 @@ pub mod pallet {
...
-    /// Block storage for accepted currency price
+    /// Asset prices from the spot price provider or the fallback price if the price is not available. Updated at the beginning of every block.
     #[pallet::storage]
     #[pallet::getter(fn currency_price)]
     pub type AcceptedCurrencyPrice<T: Config> = StorageMap<_, Twox64Concat, AssetIdOf<T>, Price, OptionQuery>;
 
-    /// Account to use when pool does not exist.
-    #[pallet::storage]
-    #[pallet::getter(fn fallback_account)]
-    pub type FallbackAccount<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
```
enthusiastmartin, line 207: No need for migration to delete the storage ? only if upgraded after it is already in use.

transaction-multi-payment/src/traits.rs
```diff
@@ -1,12 +1,27 @@
...
+    Native(Balance),
+    NonNative(Balance, AssetId, Price),
+}
+
+pub trait TransactionMultiPaymentDataProvider<AccountId, AssetId, Price> {
+    fn get_currency_and_price(who: &AccountId) -> Result<(AssetId, Option<Price>), DispatchError>;
+
+    fn get_fee_receiver() -> AccountId;
+}
+
```
enthusiastmartin, line 14: Some description would be helpful.

transaction-multi-payment/src/traits.rs
```diff
@@ -1,12 +1,27 @@
...
-    fn withdraw(
-        who: &AccountId,
-        fee: Balance,
-    ) -> Result<PaymentWithdrawResult, frame_support::sp_runtime::DispatchError>;
+    fn withdraw(who: &AccountId, fee: Balance) -> Result<PaymentWithdrawResult, DispatchError>;
+}
+
+pub trait DepositFee<AccountId, AssetId, Balance> {
+    fn deposit_fee(who: &AccountId, amounts: impl Iterator<Item = (AssetId, Balance)>) -> DispatchResult;
 }
```
enthusiastmartin, line 27: same here - description would be helpful.

enthusiastmartin (commented):
Looks good. 

we just need to make sure that no tokens are lost ( after the withdraw and deposit_fee).

transaction-multi-payment/src/lib.rs
```diff
@@ -361,68 +386,161 @@ where
...
+    fn get_fee_receiver() -> <T as frame_system::Config>::AccountId {
+        T::FeeReceiver::get()
     }
+}
 
-    fn check_balance(account: &T::AccountId, currency: AssetIdOf<T>) -> Result<(), Error<T>> {
-        if T::Currencies::free_balance(currency, account) == BalanceOf::<T>::zero() {
-            return Err(Error::<T>::ZeroBalance);
-        };
+pub struct DepositAll<T>(PhantomData<T>);
```
apopiak, line 410: docs?

transaction-multi-payment/src/lib.rs
```diff
@@ -61,17 +61,17 @@ use frame_support::traits::IsSubType;
 
 use scale_info::TypeInfo;
 
-use crate::traits::{CurrencyWithdraw, PaymentWithdrawResult};
+use crate::traits::*;
```
apopiak, line 64:
```suggestion
pub use crate::traits::*;
```
these should be public, otherwise other crates cannot use them

apopiak:
you'll also want https://github.com/galacticcouncil/warehouse/pull/48/commits/1de7570a2ae17ff088e28bd4d18144f37747a9e3

apopiak:
and to add the feature `"use_alloc"`

apopiak:
As you remove the `MultiCurrencyAdapter` this will need a companion PR for Basilisk, no?

transaction-multi-payment/src/lib.rs
Roznovjak: I wanted to make it generic (the interface is similar to [OnUnbalanced](https://docs.rs/frame-support/3.0.0/frame_support/traits/trait.OnUnbalanced.html) which was used in the original implementation).

transaction-multi-payment/src/lib.rs
Roznovjak: updated. the check for zero value is not necessary because `deposit_fee` is no-op if `total_fee == 0`

Roznovjak:
> 

Yes. We will also need another PR for `hydra-parachain`.

Roznovjak:
> As you remove the `MultiCurrencyAdapter` this will need a companion PR for Basilisk, no?

Yes. We will also need another PR for hydra-parachain.

jak-pan:
Is this ready for merge?

apopiak (approved):

enthusiastmartin:
> Is this ready for merge?

I would like to resolve first this [question ](https://github.com/galacticcouncil/warehouse/pull/48#discussion_r884218460)

It just looks to me bit unnecessary. i would simplify it.  Let's wait for Richard to give his opinion. 

Other than that - it is ready for merge.

apopiak:
He removed the iterator

Issue #45, opened 2022-05-02
feat: pallet-nft - implement non-fungibles traits.


enthusiastmartin:
pallet-nft needs to implement non-fungibles traits from support library.

It is required by eg.omnipool pallet, but i believe it will be required in xcm too when nft are transferable.

all traits from here:

https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/nonfungibles.rs

probably these too:
https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/nonfungible.rs

PR #46, opened 2022-05-05
feat: NFT implements non-fungible traits

Roznovjak:
Fixes: #45

### DIFF
--- a/nft/src/lib.rs
+++ b/nft/src/lib.rs
@@ -497,7 +497,7 @@ impl<T: Config> Destroy<T::AccountId> for Pallet<T> {
         ensure!(T::Permissions::can_destroy(&class_type), Error::<T>::NotPermitted);
         ensure!(witness.instances == 0u32, Error::<T>::TokenClassNotEmpty);
 
-        let owner = pallet_uniques::Pallet::<T>::class_owner(&class.into()).ok_or(Error::<T>::ClassUnknown)?;
+        let owner = Self::class_owner(class).ok_or(Error::<T>::ClassUnknown)?;
 
         pallet_uniques::Pallet::<T>::do_destroy_class(class.into(), witness, maybe_check_owner)?;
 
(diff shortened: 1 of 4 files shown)


nft/src/lib.rs
```diff
@@ -415,3 +415,115 @@ impl<T: Config> Pallet<T> {
...
+    fn classes() -> Box<dyn Iterator<Item = Self::ClassId>> {
+        Box::new(Classes::<T>::iter_keys())
+    }
+
+    fn instances(class: &Self::ClassId) -> Box<dyn Iterator<Item = Self::InstanceId>> {
+        Box::new(Instances::<T>::iter_key_prefix(class))
+    }
+
+    fn owned(who: &T::AccountId) -> Box<dyn Iterator<Item = (Self::ClassId, Self::InstanceId)>> {
+        Box::new(pallet_uniques::Pallet::<T>::owned(who).map(|i| (i.0.into(), i.1.into())))
```
lumir-mrkva, line 455: nitpick: would be nicer to deconstruct tuple instead of using generic indexes `.0` & `.1`

Roznovjak: Thx. Updated.

Roznovjak:
@enthusiastmartin Do we want to also implement `Mutate` trait?

enthusiastmartin:
> @enthusiastmartin Do we want to also implement `Mutate` trait? Right now we are using default implementation.

i believe so. 

the default impl returns an Unsupported error.

i think it should call do_mint ?!

nft/src/lib.rs
```diff
@@ -415,3 +415,115 @@ impl<T: Config> Pallet<T> {
...
+
+        let owner = pallet_uniques::Pallet::<T>::class_owner(&class.into()).ok_or(Error::<T>::ClassUnknown)?;
+
+        pallet_uniques::Pallet::<T>::do_destroy_class(class.into(), witness, maybe_check_owner)?;
+
+        Classes::<T>::remove(class);
+
+        Self::deposit_event(Event::ClassDestroyed { owner, class_id: class });
+
+        Ok(witness)
```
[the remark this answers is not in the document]
Roznovjak: Yes, it's there, but it doesn't return `DestroyWitness` struct required in the return value and I didn't want to fetch it two times.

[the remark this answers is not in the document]
Roznovjak: refactored

nft/src/lib.rs
```diff
@@ -412,6 +413,133 @@ impl<T: Config> Pallet<T> {
...
+
+impl<T: Config> Mutate<T::AccountId> for Pallet<T> {
+    fn mint_into(class: &Self::ClassId, instance: &Self::InstanceId, who: &T::AccountId) -> DispatchResult {
+       let class_type = Self::classes(class)
+                .map(|c| c.class_type)
+                .ok_or(Error::<T>::ClassUnknown)?;
+
+        ensure!(T::Permissions::can_mint(&class_type), Error::<T>::NotPermitted);
+
+        Self::do_mint(who.clone(), *class, *instance, BoundedVec::default())?;
```
enthusiastmartin, line 513: @green-jay Is there a reason why the permission cannot be inside the do_mint ? 

Since it is public functions and it is used by liquidity mining - i;d expect to check the permission within the do_mint.

asking because this implemenation is basically the same as mint dispatchable, and i think the permission check should be part of the do_mint. But perhaps, there is a reason why you did not do it that way.

same applies, for all do_* methods.

green-jay: it was a concious design choice, my thinking was: there will be extrinsics where all the checks will be performed, such as origin, class permissions etc. and then there will be interface functions where implementor should provide the safeguards but by default there will be none. But if it makes life easier, can be probably moved inside of interface functions as well. No strong opinions there.

green-jay (approved):
lgtm (except formatting)

PR #48, opened 2022-05-13
feat: Add custom XCM Weight Trader

apopiak:
This PR adds:
+ a `hydradx-adapters` crate to the warehouse that (for now) contains:
  + the custom XCM weight trader `MultiCurrencyTrader`
  + the struct `ToFeeReceiver` that implements `TakeRevenue` and channels it to pallet transaction multi payment's fee receiver
+ a `NativePriceOracle` trait in `hydradx-traits` and an implementation in pallet multi-transaction-payment

based on https://github.com/galacticcouncil/warehouse/pull/42

**Todo:** 
+ discarded the idea of a custom `WeightToFee`
<del>move from `trait PriceOracle` to a custom `WeightToFee` trait that can be parameterized on currencies.</del>


adapters/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+use polkadot_xcm::latest::prelude::*;
+use sp_runtime::{
+    traits::{Convert, Saturating, Zero},
+    FixedPointNumber, FixedPointOperand, FixedU128, SaturatedConversion,
+};
+use sp_std::{collections::btree_map::BTreeMap, marker::PhantomData};
+use xcm_builder::TakeRevenue;
+use xcm_executor::{traits::WeightTrader, Assets};
+
+pub type Price = FixedU128;
```
apopiak, line 31: Just set this type for now, should this be part of the generics for the trader?

traits/src/lib.rs
```diff
@@ -213,3 +213,9 @@ pub trait CanCreatePool<AssetId> {
 pub trait LockedBalance<AssetId, AccountId, Balance> {
     fn get_by_lock(lock_id: LockIdentifier, currency_id: AssetId, who: AccountId) -> Balance;
 }
+
+/// Very simple price oracle trait.
+/// TODO: Properly define the price oracle interface.
+pub trait PriceOracle<AssetId, Price> {
```
apopiak, line 219: Please comment on ways you would want the trait to be different.

adapters/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    macro_rules! generate_revenue_type {
+        ($name:ident) => {
+            lazy_static::lazy_static! {
+                pub static ref TAKEN: Mutex<BTreeSet<MultiAsset>> = Mutex::new(BTreeSet::new());
+                pub static ref EXPECTED: Mutex<BTreeSet<MultiAsset>> = Mutex::new(BTreeSet::new());
+            };
+
+            struct $name;
```
apopiak, line 252: Got races when adding this as regular global state due to cargo running tests in parallel. Threw together this hacky one type per test thing to get it done, suggestions welcome.

adapters/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+    /// Will refund up to `weight` from the first asset tracked by the trader.
+    fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
+        log::trace!(
+            target: "xcm::weight", "MultiCurrencyTrader::refund_weight weight: {:?}, paid_assets: {:?}",
+            weight, self.paid_assets
+        );
+        let weight = weight.min(self.weight);
+        self.weight -= weight; // Will not underflow because of `min()` above.
+        let fee = WeightToFee::calc(&weight);
+        if let Some(((asset_loc, price), amount)) = self.paid_assets.iter_mut().next() {
```
Roznovjak, line 134: Is it ok that the iterator we get is sorted by key?

apopiak: I don't see any issues as refund is best effort. Do you see any?

adapters/src/lib.rs
enthusiastmartin: What i usually do - i reset this to default value at the beginning of each test. 

eg. when i initialize ExtBuilder and build Test runtime.

adapters/src/lib.rs
enthusiastmartin: Is Balance in the trader generic ?

enthusiastmartin: ah I see it now. it is.

enthusiastmartin: there is a need for balance to be possible to convert to u128. 

I wonder if we can just use u128 directly for Balance too?

adapters/src/lib.rs
```diff
@@ -0,0 +1,545 @@
...
+        WeightToFee: WeightToFeePolynomial<Balance = Balance>,
+        AcceptedCurrencyPrices: PriceOracle<AssetId, Price>,
+        ConvertCurrency: Convert<MultiAsset, Option<AssetId>>,
+        Revenue: TakeRevenue,
+    > WeightTrader
+    for MultiCurrencyTrader<AssetId, Balance, WeightToFee, AcceptedCurrencyPrices, ConvertCurrency, Revenue>
+{
+    fn new() -> Self {
+        Self {
+            weight: 0,
```
enthusiastmartin, line 95: nitty picky
```suggestion
            weight: Default::default(),
```

adapters/src/lib.rs
apopiak: Ah, yes, I think that's the thing I forgot and was trying to remember :+1: :pray:

transaction-multi-payment/src/traits.rs
```diff
@@ -1,12 +1,27 @@
...
 pub trait CurrencyWithdraw<AccountId, Balance> {
-    fn withdraw(
-        who: &AccountId,
-        fee: Balance,
-    ) -> Result<PaymentWithdrawResult, frame_support::sp_runtime::DispatchError>;
+    fn withdraw(who: &AccountId, fee: Balance) -> Result<PaymentWithdrawResult, DispatchError>;
+}
+
+pub trait DepositFee<AccountId, AssetId, Balance> {
+    fn deposit_fee(who: &AccountId, amounts: impl Iterator<Item = (AssetId, Balance)>) -> DispatchResult;
```
enthusiastmartin, line 26: i commented in the other [PR](https://github.com/galacticcouncil/warehouse/pull/42#discussion_r883565573)

What is the need to have the iterator ? 

it seems that only one value is always used. 

i wonder whether params  (who, currency, amount) would not sufficient.

adapters/src/lib.rs
Roznovjak: No, I don't see any issue. I just wanted to be sure that we are aware of it.

transaction-multi-payment/src/traits.rs
apopiak: :ok_hand: I just based this PR on the tx fee one. I agree we might not need the iterator, but would adjust that in the original PR.

enthusiastmartin: @Roznovjak  wdyt ?

Roznovjak: the iterator was removed in the original PR.

apopiak: He adjusted the PR to remove the iterator.

adapters/src/lib.rs
apopiak: Sure we could, would limit the type to runtimes with `type Balance = u128`, though

apopiak: I think it would be nice to make it somewhat reusable. Maybe I should even make Price be a generic `FixedPointNumber` :thinking:

traits/src/lib.rs
enthusiastmartin: we already have SpotPriceProvider trait. 

This returns a price for given asset pair. 

But i dont think this is a correct name for this trait . It is not really a price oracle. Because as i can understand - it used to determine whether given currecny is also accepted fee currency, isn't it ?

adapters/src/lib.rs
apopiak: Right, this being in the warehouse which are meant to be more generic building blocks I guess it would make sense to make a different tradeoff?

adapters/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+impl<
+        AssetId,
+        Balance: FixedPointOperand + TryInto<u128>,
+        WeightToFee: WeightToFeePolynomial<Balance = Balance>,
+        AcceptedCurrencyPrices: PriceOracle<AssetId, Price>,
+        ConvertCurrency: Convert<MultiAsset, Option<AssetId>>,
+        Revenue: TakeRevenue,
+    > Drop for MultiCurrencyTrader<AssetId, Balance, WeightToFee, AcceptedCurrencyPrices, ConvertCurrency, Revenue>
```
enthusiastmartin, line 162: Could you add a short comment about how this works, how/when it used.

it is clear to me but it might be useful for newcomers.

traits/src/lib.rs
apopiak: Mmmh, it's "give me a trusted price for this asset", which in this case is determined by whether it's an accepted fee currency, but that seems an implementation detail to me.

enthusiastmartin (commented):
looks ok me. 

you can actually merge the latest master - it will simplify as some of the changes were done as part of another Pr

adapters/src/lib.rs
apopiak: Done. I'm a Rust god now, btw. it compiled and ran tests successfully on the first try :1st_place_medal:

traits/src/lib.rs
enthusiastmartin: actually - this is ok.

i wonder if we can do something like this :

```rust
trait PriceOracle<Assets, Price> {
     fn price(currency: &Assets) -> Option<Price>;
}
```

My intention here is to express that Assets can be just t one asset or asset pair.

in the trader you can require Assets to be just AssetId - one asset id like you do now.

But we can use this in similar way how we use SpotPriceProvider, where we set Assets = AssetPair.

actually, we can unite this into one trait then? wdyt ?

enthusiastmartin: perhaps. the thing is that me, i'd expect a price oracle to give me a price of any asset pair if possible.

that;s why i see it as two traits doing same thing, except here you know that you want a price for native/target currency pair.

so, have a separate trait for such case is ok, but i'd prefer different name, eg  "NativePriceOracle".  the "trusted" prefix is not needed imo.

enthusiastmartin (approved):
There is still some issue in the build ( clippy is complaining i believe)

lgtm when fixed and build passes

apopiak:
> There is still some issue in the build ( clippy is complaining i believe)
> 
> lgtm when fixed and build passes

I'm unsure how to fix the clippy error as it doesn't seem to be in our repo but in Polkadot?

enthusiastmartin:
> > There is still some issue in the build ( clippy is complaining i believe)
> > lgtm when fixed and build passes
> 
> I'm unsure how to fix the clippy error as it doesn't seem to be in our repo but in Polkadot?

Fixed it for you for now.

it is due to the runtime-benchmarks features used in the repo. 

and we run clippy with all-features.

PR #50, opened 2022-05-24
Feat: liquidity mining

martinfridrich:
https://github.com/galacticcouncil/Basilisk-node/issues/446


liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+    /// pools and can be distributed in a longer time frame but never in the shorter time frame.
+    /// - `blocks_per_period`:  number of blocks in a single period. Min. number of blocks per
+    /// period is 1.
+    /// - `incentivized_asset`: asset to be incentivized in AMM pools. All liq. pools added into
+    /// liq. mining program have to have `incentivized_asset` in their pair.
+    /// - `reward_currency`: payoff currency of rewards.
+    /// - `owner`: liq. mining farm owner.
+    /// - `yield_per_period`: percentage return on `reward_currency` of all pools
+    #[allow(clippy::too_many_arguments)]
+    #[transactional]
```
enthusiastmartin, line 287: I believe this will part of another transactional operation. 
```suggestion
    #[require_transactional]
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+        let global_pool_account = Self::pool_account_id(global_pool.id)?;
+
+        let undistributed_reward = T::MultiCurrency::total_balance(global_pool.reward_currency, &global_pool_account);
+
+        T::MultiCurrency::transfer(
+            global_pool.reward_currency,
+            &global_pool_account,
+            &who,
+            undistributed_reward,
+        )?;
```
enthusiastmartin, line 397: I believe that the transfer would fail if there is a locked/reserved amount in global_pool_account. 

because the total_balance returns combined balance but the transfer checks only the "free" part.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+            DepositId,
+            Self::AccountId,
+        >;
+    }
+
+    #[pallet::error]
+    #[cfg_attr(test, derive(PartialEq))]
+    pub enum Error<T> {
+        /// Math computation overflow.
+        Overflow,
```
enthusiastmartin, line 148: Instead of custom overflow, you can use  `sp_runtime::ArithmeticError;` 

that's what we started to do on other crates where math error is needed/

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1299 @@
...
+    #[pallet::getter(fn liquidity_pool)]
+    pub type LiquidityPoolData<T: Config> = StorageDoubleMap<
+        _,
+        Twox64Concat,
+        GlobalPoolId,
+        Twox64Concat,
+        T::AmmPoolId,
+        LiquidityPoolYieldFarm<T>,
+        OptionQuery,
+    >;
```
enthusiastmartin, line 248: it is not clear who can create T::AmmPoolIds 

so i'd suggest to use "Blake2_128Concat" hasher. Also it might preferable choice in other other storages too.

enthusiastmartin (commented):
I havent looked at it all yet. 

but here my initial findings. i'll continue again later.

PR #51, opened 2022-06-02
feat: introduce new accountidfor + reorganize registry traits

enthusiastmartin:
Replace Current "ShareAccountIdFor" with improved version where it can handle more than 2 assets.

This might be eventually needed for Stableswap with more than 2 assets in pool.

### DIFF
--- a/traits/src/lib.rs
+++ b/traits/src/lib.rs
@@ -19,6 +19,8 @@
 #![allow(clippy::upper_case_acronyms)]
 
 pub mod pools;
+pub mod registry;
+pub use registry::*;
 
 use codec::{Decode, Encode};
 use frame_support::dispatch;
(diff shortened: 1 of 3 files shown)


traits/src/lib.rs
```diff
@@ -132,48 +134,6 @@ pub trait Resolver<AccountId, Intention, E> {
...
-    fn create_shared_asset(
-        name: &AssetName,
-        assets: &[AssetId],
-        existential_deposit: Balance,
-    ) -> Result<AssetId, Error>;
-
-    fn get_or_create_shared_asset(
-        name: AssetName,
-        assets: Vec<AssetId>,
-        existential_deposit: Balance,
```
Roznovjak, line 163: It would be nice (if possible) to have the same types as in `create_shared_asset` (a reference for `AssetName` and a slice for assets).

traits/src/lib.rs
```diff
@@ -132,48 +134,6 @@ pub trait Resolver<AccountId, Intention, E> {
...
-    fn get_or_create_asset(name: AssetName, existential_deposit: Balance) -> Result<AssetId, Error> {
-        if let Ok(asset_id) = Self::retrieve_asset(&name) {
-            Ok(asset_id)
-        } else {
-            Self::create_asset(&name, existential_deposit)
-        }
-    }
-}
-
-pub trait ShareTokenRegistry<AssetId, AssetName, Balance, Error>: Registry<AssetId, AssetName, Balance, Error> {
```
Roznovjak, line 151: Could you please provide short description for this trait? Mainly when to use this one instead of `Registry`

Roznovjak (approved):

traits/src/lib.rs
enthusiastmartin: registry traits should be refactored and changed ( this was done long time ago and it was not the best )

but this PR is mainly about the new trait and moving registry traits into separate file for better readability.

that means it cannot be done now, as it would be breaking change.

traits/src/lib.rs
enthusiastmartin: It will be eventually done as part of registry traits refactor.

PR #54, opened 2022-06-06
Add types to set tx fee asset on new account

apopiak:
Add structs to the multi tx payment pallet that set and delete tx fee currency on account creation and deletion.

`AddTxAssetOnAccount` sets the tx fee currency on account creation if it is not set already.

`RemoveTxAssetOnKilled` removes the tx fee currency if the system account is removed or if the tx currency balance hits zero.

based on https://github.com/apopiak/open-runtime-module-library/pull/1 (backport of https://github.com/open-web3-stack/open-runtime-module-library/pull/754), we will thus need to maintain a fork of ORML until we're up-to-date in terms of the substrate/polkadot versions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,15 @@ members = [
 ]
 
 resolver = "2"
+
+[patch."https://github.com/open-web3-stack/open-runtime-module-library"]
+orml-currencies = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-tokens = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-traits = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-vesting = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-benchmarking = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17", optional = true }
+orml-xtokens = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-xcm-support = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-unknown-tokens = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-xcm = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-utilities = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,8 @@ members = [
 
 resolver = "2"
 
+# Use a fork of ORML until we are using a version that includes
+# https://github.com/open-web3-stack/open-runtime-module-library/pull/754
 [patch."https://github.com/open-web3-stack/open-runtime-module-library"]
 orml-currencies = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
 orml-tokens = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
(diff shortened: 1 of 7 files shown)


transaction-multi-payment/src/lib.rs
```diff
@@ -576,3 +576,27 @@ impl<T: Config + Send + Sync> CurrencyBalanceCheck<T> {
...
+            && (*currency == T::NativeAssetId::get() || AcceptedCurrencies::<T>::contains_key(&currency))
+        {
+            AccountCurrencyMap::<T>::insert(&who, currency);
+        }
+    }
+}
+
+/// Type to automatically remove the fee currency for an account on account deletion.
+///
+/// Note: The fee currency is only removed if the system account is gone.
```
jak-pan, line 594: Is this a technical limitation or logical decision? I would tend to think that we might also do this when your balance for the asset that was set was reaped. Just a thought.

Cargo.toml
```diff
@@ -13,3 +13,15 @@ members = [
...
+orml-tokens = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-traits = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-vesting = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-benchmarking = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17", optional = true }
+orml-xtokens = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-xcm-support = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-unknown-tokens = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-xcm = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
+orml-utilities = { git = "https://github.com/apopiak/open-runtime-module-library", branch = "apopiak/on-new-account-9-17" }
```
enthusiastmartin, line 27: why need of all these crates ?

transaction-multi-payment/src/lib.rs
```diff
@@ -576,3 +576,27 @@ impl<T: Config + Send + Sync> CurrencyBalanceCheck<T> {
...
+pub struct RemoveTxAssetOnKilled<T>(PhantomData<T>);
+impl<T: Config> Happened<(T::AccountId, AssetIdOf<T>)> for RemoveTxAssetOnKilled<T> {
+    fn happened((who, _currency): &(T::AccountId, AssetIdOf<T>)) {
+        if !frame_system::Pallet::<T>::account_exists(&who) {
+            AccountCurrencyMap::<T>::remove(&who);
+        }
```
enthusiastmartin, line 600: i wonder if you could just call remove without the check. 

can there be a case where you would like to keep it in the currency map if it does not exist ?

enthusiastmartin: ah..i think i misread the condition

transaction-multi-payment/src/lib.rs
apopiak: Ah, interesting idea!

transaction-multi-payment/src/lib.rs
enthusiastmartin: is the "happened" called only when orml-tokens removes account from its storage? 

i guess so, so in that case - ignore this comment.

Cargo.toml
apopiak: Because I'm using https://github.com/open-web3-stack/open-runtime-module-library/pull/754 , it's not merged and I just replaced all of the orml crates used. Happy to reduce the list if we don't need some of them.

transaction-multi-payment/src/lib.rs
apopiak: We should not just remove because this is called for *every currency account removed*.
So if you have BSX, KSM and aUSD in your account and transfer out all the aUSD, this callback will be called, but would not lead us to want to remove the tx fee currency (especially not if it's set to KSM).

transaction-multi-payment/src/lib.rs
apopiak: Ok, implemented. Note, though, that this incurs more storage reads on deletion.

enthusiastmartin (changes requested):
looks good to me.

enthusiastmartin (approved):

jak-pan (approved):
LGTM

PR #55, opened 2022-06-07
feat: liquidity mining with multiple incentives

martinfridrich:
General implementation of liquidity mining pallet with multiple incentives usable by other pallets.

https://github.com/galacticcouncil/Basilisk-node/issues/461

* https://www.notion.so/Liquidity-Mining-codename-double-trouble-2ad34089116047df8eeddba20ecdb383
* https://app.clickup.com/t/2fkj6zp


liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1401 @@
+// This file is part of HydraDX
+
+// Copyright (C) 2020-2021  Intergalactic, Limited (GIB).
```
enthusiastmartin, line 3: 2022

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1401 @@
...
+            FarmId,
+            Balance,
+            DepositId,
+            Self::AccountId,
+        >;
+
+        /// Maximum number of yield farms same LP shares can be re/deposited into. This value always
+        /// MUST BE >= 1.         
+        #[pallet::constant]
+        type MaxFarmEntriesPerDeposit: Get<u8>;
```
enthusiastmartin, line 187: What if you use Range or RangeInclusive? 

that way, you can specify min and max bounds easily. You wont need to check if it > 1, you would just check if value is within range. 

```suggestion
        type MaxFarmEntriesPerDeposit: Get<RangeInclusive<u8>>;
```

liquidity-mining/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+license = "Apache 2.0"
+repository = "https://github.com/galacticcouncil/warehouse"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+codec = { default-features = false, features = ["derive"], package = "parity-scale-codec", version = "2.3.1" }
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
+serde = { features = ["derive"], optional = true, version = "1.0.101" }
```
enthusiastmartin, line 17: serde not needed.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+            incentivized_asset,
+            max_reward_per_period,
+            state: GlobalFarmState::Active,
+        }
+    }
+
+    /// This function updates yields_farm_count when new yield farm is added into the global farm.
+    /// This function should be called only when new yield farm is created/added into the global
+    /// farm.
+    pub fn yield_farm_added(&mut self) -> Result<(), ArithmeticError> {
```
enthusiastmartin, line 76: the name of this function does not give impression that it updates some counts. 

i'd suggest to call increase_farm_counts or something similar.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+                .checked_add(1)
+                .ok_or(ArithmeticError::Overflow)?,
+        );
+
+        Ok(())
+    }
+
+    /// This function updates `yield_farms_count` when yield farm is removed from global farm.
+    /// This function should be called only when yield farm is removed from global farm.
+    pub fn yield_farm_removed(&mut self) -> Result<(), ArithmeticError> {
```
enthusiastmartin, line 93: same here, similar to preivous comment.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+            .checked_sub(1)
+            .ok_or(ArithmeticError::Underflow)?;
+
+        Ok(())
+    }
+
+    /// This function updates `yield_farms_count` when yield farm is flushed from storage.
+    /// This function should be called only if yield farm is flushed.
+    /// DON'T call this function if yield farm is in stopped or deleted state.
+    pub fn yield_farm_flushed(&mut self) -> Result<(), DispatchError> {
```
enthusiastmartin, line 107: also here, i would go with more descriptive name.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+            .yield_farms_count
+            .1
+            .checked_sub(1)
+            .ok_or(ArithmeticError::Underflow)?;
+
+        Ok(())
+    }
+
+    /// Function returns `true` if global farm has no live yield farms.
+    pub fn has_no_live_farms(&self) -> bool {
```
enthusiastmartin, line 118: personally, i would do "has_live_farms" instead.   Negative questions/conditions usually have some additional thinking involved;-)

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+pub type FarmId = u32;
+pub type GlobalFarmId = FarmId;
+pub type YieldFarmId = FarmId;
+pub type FarmMultiplier = FixedU128;
+pub type DepositId = u128;
+
+/// This type represent number of live(active and stopped)` yield farms in global farm.
+pub type LiveFarmsCount = u32;
+/// This type represent number of total(active, stopped and deleted)` yield farms in global farm.
+pub type TotalFarmsCount = u32;
```
enthusiastmartin, line 14: i don't think you need to create these two types. just use u32 directly. 

If you want to make sure that it is clear what the parts mean in the tuple (yield_farm_count) - replace the tuple with a struct, where you use live_farms as variable instead of count.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    /// Returns `true` if yield farm can be removed from storage, `false` otherwise.
+    pub fn can_be_flushed(&self) -> bool {
+        self.state == YieldFarmState::Deleted && self.entries_count.is_zero()
+    }
+
+    /// This function updates entries count in the yield farm. This function should be called if  
+    /// entry is removed from the yield farm.
+    pub fn entry_removed(&mut self) -> Result<(), ArithmeticError> {
```
enthusiastmartin, line 192: again, i would prefer more descriptive name here. It decrease a count. but the name doesnt suggest that.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+        self.entries_count = self.entries_count.checked_sub(1).ok_or(ArithmeticError::Underflow)?;
+
+        Ok(())
+    }
+
+    /// This function updates entries count in the yield farm. This function should be called if
+    /// entry is added into the yield farm.
+    pub fn entry_added(&mut self) -> Result<(), ArithmeticError> {
```
enthusiastmartin, line 200: i'd just do "increase_entries".

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+        self.entries_count = self.entries_count.checked_add(1).ok_or(ArithmeticError::Overflow)?;
+
+        Ok(())
+    }
+
+    /// This function return `true` if yield farm is empty.
+    pub fn has_entries(&self) -> bool {
```
enthusiastmartin, line 207: This is what i meant in the previous comments about "has_no_live_farms" .

Here you could do the same " has_no_entries"  but decided to do the other way;-)  I belive this is a better way.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// This function return yield farm entry from deposit of `None` if yield farm entry is not
+    /// found.
+    pub fn get_yield_farm_entry(&mut self, yield_farm_id: FarmId) -> Option<&mut YieldFarmEntry<T>> {
+        match self
+            .yield_farm_entries
+            .binary_search_by(|e| e.yield_farm_id.cmp(&yield_farm_id))
+        {
+            Ok(idx) => self.yield_farm_entries.get_mut(idx),
+            Err(_) => None,
+        }
```
enthusiastmartin, line 290: see previous comment.

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+    Active,
+    Deleted,
+}
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebugNoBound, TypeInfo)]
+pub enum YieldFarmState {
+    Active,
+    Stopped,
+    Deleted,
+}
```
enthusiastmartin, line 354: This could not be merged into one ?  

I see there is one extra state in yield farm state, but if you ensure such state cannot be set for global, you should be fine.

traits/src/liquidity_mining.rs
```diff
@@ -0,0 +1,33 @@
+use frame_support::dispatch::DispatchError;
+
+pub trait Handler<AssetId, AmmPoolId, GlobaFarmId, YieldFarmId, Balance, DepositId, AccountId> {
```
enthusiastmartin, line 3: I already mentioned this in my previous review. 

More descriptive trait name would be better. Handler is very generic imo.

traits/src/liquidity_mining.rs
```diff
@@ -0,0 +1,33 @@
...
+        total_valued_shares: Balance,
+    );
+
+    /// This function should lock LP shares.
+    fn lock_lp_tokens(
+        amm_pool_id: AmmPoolId,
+        who: AccountId,
+        amount: Balance,
+        deposit_id: DepositId,
+    ) -> Result<(), DispatchError>;
```
enthusiastmartin, line 24: i;d suggest to create associated type error in this trait instead of specific substrate DispatchError. 

that menas you add:

type Error; 

and then use it in:

Result<(),Self::Error>

enthusiastmartin (commented):
initial findings. 

I'll continue later with the lib.rs

liquidity-mining/Cargo.toml
```diff
@@ -0,0 +1,55 @@
+[package]
+name = "pallet-liquidity-mining"
+version = "1.0.0"
```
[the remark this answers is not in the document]
martinfridrich: When I used `0.1.0` in different PR @lumir-mrkva suggested to use `1.0.0`

liquidity-mining/src/types.rs
martinfridrich: Yes, I decided to name them based on how this fn are used(they are never called with `!`) but I'll change it

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Encode, Decode, RuntimeDebugNoBound, TypeInfo, PartialEq)]
+pub struct DepositData<T: Config> {
+    pub shares: Balance,
+    pub amm_pool_id: T::AmmPoolId,
+    //NOTE: Capacity of this vector MUST BE at least 1.
+    pub yield_farm_entries: Vec<YieldFarmEntry<T>>,
```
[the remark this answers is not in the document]
martinfridrich: Yes, I thought about using `BoundedVec`, I'll look at it

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,354 @@
...
+            //NOTE: Capacity of this vector MUST BE at least 1.
+            yield_farm_entries: vec![],
+        }
+    }
+
+    /// This function add new yield farm entry into the deposit.
+    /// This function returns error if deposit reached max entries in the deposit or
+    /// `entry.yield_farm_id` is not unique.
+    pub fn add_yield_farm_entry(&mut self, entry: YieldFarmEntry<T>) -> Result<(), DispatchError> {
+        let len = TryInto::<u8>::try_into(self.yield_farm_entries.len()).map_err(|_e| ArithmeticError::Overflow)?;
```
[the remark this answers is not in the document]
martinfridrich: yes, vec is timited by `T::MaxFarmEntriesPerDeposit`. We assume it will be small number e.g 2-4, max 5.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1415 @@
...
+        InvalidInitialRewardPercentage,
+
+        /// One or more yield farms exist in global farm.
+        GlobalFarmIsNotEmpty,
+
+        /// Farm's `incentivized_asset` is missing in provided asset pair.
+        MissingIncentivizedAsset,
+
+        /// Global's farm rewards balance is not 0.
+        RewardBalanceIsNotZero,
```
dmoka, line 228: I think that this `RewardBalanceIsNotZero` can be removed as i have not found any usage in this file.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+            min_deposit,
+        )?;
+
+        ensure!(
+            T::MultiCurrency::free_balance(reward_currency, &owner) >= total_rewards,
+            Error::<T, I>::InsufficientRewardCurrencyBalance
+        );
+
+        let planned_periods =
+            TryInto::<u128>::try_into(planned_yielding_periods).map_err(|_e| ArithmeticError::Overflow)?;
```
enthusiastmartin, line 356: Also, i would probably choose different error - there is already something like InvalidPeriod. 

```suggestion
            TryInto::<u128>::try_into(planned_yielding_periods).map_err(|_| ArithmeticError::Overflow)?;
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+
+            T::MultiCurrency::transfer(
+                global_farm.reward_currency,
+                &global_farm_account,
+                &who,
+                undistributed_rewards,
+            )?;
+
+            //Mark for flush from storage on last `YieldFarm` in the farm flush.
+            global_farm.state = FarmState::Deleted;
```
enthusiastmartin, line 420: is this necessary ? you already checked "has_live_farms", so it could be flushed? or there can be scenarios where it cannot be done?

the thing is that name is "destroy_global_farm". so i would expect the global farm to be destroyed ( like remove from storage) .

but it seems that this returns OK but in some cases the global farm is not destroyed, but only state is updated. 

Is there then another way to destroy the form for sure ?

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+
+            Ok((reward_currency, undistributed_rewards, who))
+        })
+    }
+
+    /// Add yield farm to global farm and start liquidity mining for given assets pair.
+    ///  
+    /// Only farm owner can perform this action.
+    ///
+    /// One of the AMM assets HAVE to be `incentivized_token`. Same AMM can be
```
enthusiastmartin, line 435:
```suggestion
    /// One of the AMM assets has to be `incentivized_token`. Same AMM can be
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+            );
+        }
+
+        <GlobalFarm<T, I>>::try_mutate(
+            global_farm_id,
+            |maybe_global_farm| -> Result<YieldFarmId, DispatchError> {
+                let global_farm = maybe_global_farm.as_mut().ok_or(Error::<T, I>::GlobalFarmNotFound)?;
+
+                //This is basically same as farm not found.
+                ensure!(global_farm.is_active(), Error::<T, I>::GlobalFarmNotFound);
```
enthusiastmartin, line 471: Why not return GlobalFarmCanceled?

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+
+                    // update global farm accumulated RPZ
+                    let current_period = Self::get_current_period(global_farm.blocks_per_period)?;
+                    if !global_farm.total_shares_z.is_zero() && global_farm.updated_at != current_period {
+                        let reward_per_period = math::calculate_global_pool_reward_per_period(
+                            global_farm.yield_per_period.into(),
+                            global_farm.total_shares_z,
+                            global_farm.max_reward_per_period,
+                        )
+                        .map_err(|_e| ArithmeticError::Overflow)?;
```
enthusiastmartin, line 493:
```suggestion
                        .map_err(|_| ArithmeticError::Overflow)?;
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+                    Error::<T, I>::MissingIncentivizedAsset
+                );
+
+                <ActiveYieldFarm<T, I>>::try_mutate(amm_pool_id.clone(), &global_farm_id, |maybe_active_yield_farm| {
+                    ensure!(maybe_active_yield_farm.is_none(), Error::<T, I>::YieldFarmAlreadyExists);
+
+                    // update global farm accumulated RPZ
+                    let current_period = Self::get_current_period(global_farm.blocks_per_period)?;
+                    if !global_farm.total_shares_z.is_zero() && global_farm.updated_at != current_period {
+                        let reward_per_period = math::calculate_global_pool_reward_per_period(
```
enthusiastmartin, line 488: I believe this function in the math could return just Option<Balance>. But this is not in this PR.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+    /// Parameters:
+    /// - `who`: farm's owner
+    /// - `global_farm_id`: global farm id in which yield farm will be updated.
+    /// - `multiplier`: new yield farm multiplier.
+    /// - `amm_pool_id`: identifier of the AMM pool.
+    pub fn update_yield_farm_multiplier(
+        who: AccountIdOf<T>,
+        global_farm_id: GlobalFarmId,
+        multiplier: FarmMultiplier,
+        amm_pool_id: T::AmmPoolId,
```
enthusiastmartin, line 526: nitpick. i would swap these two params ( the reason beging that first you specify what/where to update and then the new value).

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+        ensure!(!multiplier.is_zero(), Error::<T, I>::InvalidMultiplier);
+
+        let yield_farm_id =
+            Self::active_yield_farm(amm_pool_id.clone(), global_farm_id).ok_or(Error::<T, I>::YieldFarmNotFound)?;
+
+        <YieldFarm<T, I>>::try_mutate((amm_pool_id, global_farm_id, yield_farm_id), |maybe_yield_farm| {
+            let yield_farm = maybe_yield_farm.as_mut().ok_or(Error::<T, I>::YieldFarmNotFound)?;
+
+            //This should never fail. If farm is in the `ActiveYieldFarm` storage, it MUST be
+            //active.
```
enthusiastmartin, line 537: i dont think that "this should never fail" is needed to tell here.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+                let old_stake_in_global_farm =
+                    math::calculate_global_pool_shares(yield_farm.total_valued_shares, yield_farm.multiplier)
+                        .map_err(|_e| ArithmeticError::Overflow)?;
+
+                let current_period = Self::get_current_period(global_farm.blocks_per_period)?;
+                Self::maybe_update_farms(global_farm, yield_farm, current_period)?;
+
+                let new_stake_in_global_farm =
+                    math::calculate_global_pool_shares(yield_farm.total_valued_shares, multiplier)
+                        .map_err(|_e| ArithmeticError::Overflow)?;
```
enthusiastmartin, line 554:
```suggestion
                        .map_err(|_| ArithmeticError::Overflow)?;
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+                let yield_farm_id = maybe_active_yield_farm_id
+                    .as_ref()
+                    .ok_or(Error::<T, I>::YieldFarmNotFound)?;
+
+                <YieldFarm<T, I>>::try_mutate(
+                    (amm_pool_id, global_farm_id, yield_farm_id),
+                    |maybe_yield_farm| -> Result<(), DispatchError> {
+                        let yield_farm = maybe_yield_farm.as_mut().ok_or(Error::<T, I>::YieldFarmNotFound)?;
+
+                        //NOTE: this should never fail because yield farm MUST be in the
```
enthusiastmartin, line 601: again, i donl think this is needed to be told here.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+                            ensure!(global_farm.owner == who, Error::<T, I>::Forbidden);
+
+                            let current_period = Self::get_current_period(global_farm.blocks_per_period)?;
+                            Self::maybe_update_farms(global_farm, yield_farm, current_period)?;
+
+                            let old_stake_in_global_pool = math::calculate_global_pool_shares(
+                                yield_farm.total_valued_shares,
+                                yield_farm.multiplier,
+                            )
+                            .map_err(|_e| ArithmeticError::Overflow)?;
```
enthusiastmartin, line 617:
```suggestion
                            .map_err(|_| ArithmeticError::Overflow)?;
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+
+                            global_farm.total_shares_z = global_farm
+                                .total_shares_z
+                                .checked_sub(old_stake_in_global_pool)
+                                .ok_or(ArithmeticError::Overflow)?;
+
+                            yield_farm.state = FarmState::Stopped;
+                            yield_farm.multiplier = 0.into();
```
enthusiastmartin, line 625: nitpick
```suggestion
                            yield_farm.multiplier = FarmMultiplier::default();
                            
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+            },
+        )
+    }
+
+    /// Resume liquidity mining for stopped yield farm.
+    ///
+    /// This function resume incentivization from `GlobalPool` and restore full functionality
+    /// for yield farm. Users will be able to deposit, claim and withdraw again.
+    ///
+    /// WARN: Yield farm is NOT rewarded for time it was stopped.
```
enthusiastmartin, line 646:
```suggestion
    /// Yield farm is not rewarded for the time it was stopped.
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+
+                    //update `GlobalFarm` accumulated_rpz
+                    let current_period = Self::get_current_period(global_farm.blocks_per_period)?;
+                    if !global_farm.total_shares_z.is_zero() && global_farm.updated_at != current_period {
+                        let reward_per_period = math::calculate_global_pool_reward_per_period(
+                            global_farm.yield_per_period.into(),
+                            global_farm.total_shares_z,
+                            global_farm.max_reward_per_period,
+                        )
+                        .map_err(|_e| ArithmeticError::Overflow)?;
```
enthusiastmartin, line 690:
```suggestion
                        .map_err(|_| ArithmeticError::Overflow)?;
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+                        Self::update_global_farm(global_farm, current_period, reward_per_period)?;
+                    }
+
+                    let new_stake_in_global_farm =
+                        math::calculate_global_pool_shares(yield_farm.total_valued_shares, multiplier)
+                            .map_err(|_e| ArithmeticError::Overflow)?;
```
enthusiastmartin, line 696:
```suggestion
                            .map_err(|_| ArithmeticError::Overflow)?;
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+                    //add yield farm to active farms.
+                    *maybe_active_yield_farm_id = Some(yield_farm.id);
+
+                    Ok(yield_farm.id)
+                })
+            })
+        })
+    }
+
+    /// This function mark yield farm ready for removal from storage when it's empty. Users will
```
enthusiastmartin, line 717:
```suggestion
    /// This function marks an yield farm ready for removal from storage when it's empty. Users will
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+
+                    Ok(())
+                },
+            )?;
+
+            //NOTE: This never happen. `GlobalFarm` must be active to have non-destroyed yield
+            //farms.
+            if global_farm.can_be_flushed() {
+                *maybe_global_farm = None;
+            }
```
enthusiastmartin, line 787: if this never happens - it can be deleted ?!

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+
+            Ok(())
+        })
+    }
+
+    /// Deposit LP shares to a yield farm.
+    ///
+    /// This function create new deposit farm entry in the yield farm.
```
enthusiastmartin, line 795:
```suggestion
    /// This function creates new deposit farm entry in the yield farm.
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1416 @@
...
+    pub fn deposit_lp_shares(
+        who: AccountIdOf<T>,
+        global_farm_id: GlobalFarmId,
+        yield_farm_id: YieldFarmId,
+        amm_pool_id: T::AmmPoolId,
+        shares_amount: Balance,
+    ) -> Result<DepositId, DispatchError> {
+        let mut deposit = DepositData::new(shares_amount, amm_pool_id.clone());
+
+        Self::do_deposit_lp_shares(&mut deposit, global_farm_id, yield_farm_id)?;
```
enthusiastmartin, line 813: why the "do_deposit" function cannot return the depositdata directly instead of mutating existing one ? 

it looks to me a better "functional" approach.

enthusiastmartin (commented):
I am only in the middle of it all. 

findings so far.

liquidity-mining/src/lib.rs
martinfridrich: `has_live_farms()` only checks if farm can be destroyed, farm with live farms can't be destroyed.

Yes, a farm is not flushed from storage if there is some "reference" to it. e.g deleted yield farm with deposits, that's why state update is necessary.

Farm is removed from storage with last yield farm removal from storage and yield farm is removed from storage with last deposit removal.

As a user you can't say "remove from storage". Storage clean up is "automatic". 

I used "destroy" because from the POW of farm's owner it's basically gone, it can't be resumed, he can't do anything with it...

But if you think "delete" is better I can rename it. I just wanted to emphasize that this is "destruction" of the farm.

liquidity-mining/src/lib.rs
martinfridrich: Global farm can't be canceled. It can be only `active` and `deleted` and `deleted` state is same as "not found in the storage".

liquidity-mining/src/lib.rs
martinfridrich: bacause of `redeposit_lp_shares()` - `do_deposit` should not always create new deposit, sometimes it's updateting existing deposit.

liquidity-mining/src/lib.rs
martinfridrich: It just calculate value or fail, why do you think Option better?

liquidity-mining/Cargo.toml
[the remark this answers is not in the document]
lumir-mrkva: yeah, no reason start with 0 version ever in my opinion. if you wanted to count versions from 0, do atleast 0.0.0 :)

according to semver, 0 major version should be used only when API is not stable yet, which shouldn't be case for us https://semver.org/#spec-item-4

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1616 @@
...
+// * total_shares_z         -> Z
+// * multiplier             -> m
+
+//! # Liquidity mining pallet
+//!
+//! ## Overview
+//!
+//! This pallet provide functionality for liquidity mining program with time incentive(loyalty
+//! factor) and multiple incentives scheme.
+//! Users are rewarded for each period they stay in liq. mining program.
```
apopiak, line 35:
```suggestion
//! This pallet provides functionality for a liquidity mining program with a time incentive (loyalty
//! factor) and multiple incentives scheme.
//! Users are rewarded for each period they stay in the liq. mining program.
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1616 @@
...
+//!
+//! Reward per one period is derived from the user's loyalty factor which grows with time(periods)
+//! the user is in the liq. mining and amount of LP shares user locked into deposit.
+//! User's loyalty factor is reset if the user exits and reenters liquidity mining.
+//! User can claim rewards without resetting loyalty factor, only withdrawing shares
+//! is penalized by loyalty factor reset.
+//! User is rewarded from the next period after he enters.
```
apopiak, line 42:
```suggestion
//! Reward per one period is derived from the user's loyalty factor which grows with time
//! (periods) the user is in the liq. mining program and the amount of LP shares the user locked into deposit.
//! User's loyalty factor is reset if the user exits and reenters liquidity mining.
//! User can claim rewards without resetting loyalty factor, only withdrawing shares
//! is penalized by loyalty factor reset.
//! The user is rewarded from the next period after they enter.
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1616 @@
...
+//!     * `Stopped` - liquidity mining is stopped. Users can claim and withdraw LP shares from the
+//!     farm. Users CAN'T deposit new LP shares to stopped farm. Stopped farm is not rewarded from the
+//!     `GlobalFarm`.
+//!     Note: stopped farm can be resumed or destroyed.
+//!     * `Deleted` - liquidity mining is ended. User's CAN'T deposit or claim rewards from
+//!     stopped farm. Users CAN only withdraw LP shares(without rewards).
+//!     `YiedlFarm` must be stopped before it can be deleted. Deleted farm stays in the storage
+//!     until last farm's entry is withdrawn. Last withdrawn from yield farm will remove deleted
+//!     farm from the storage.
+//!     Note: Deleted farm CAN'T be resumed.
```
apopiak, line 73:
```suggestion
//! * `YieldFarm` -  can be in the 3 states: [`Active`, `Stopped`, `Canceled`]
//!     * `Active` - liquidity mining is running, users are able to deposit, claim and withdraw LP
//!     shares. `YieldFarm` is rewarded from `GlobalFarm` in this state.
//!     * `Stopped` - liquidity mining is stopped. Users can claim and withdraw LP shares from the
//!     farm. Users CAN'T deposit new LP shares to stopped farm. Stopped farm is not rewarded from the
//!     `GlobalFarm`.
//!     Note: stopped farm can be resumed or destroyed.
//!     * `Deleted` - liquidity mining is ended. User's CAN'T deposit or claim rewards from
//!     stopped farm. Users CAN only withdraw LP shares(without rewards).
//!     `YieldFarm` must be stopped before it can be deleted. Deleted farm stays in the storage
//!     until last farm's entry is withdrawn. Last withdrawn from yield farm will remove deleted
//!     farm from the storage.
//!     Note: Deleted farm CAN'T be resumed.
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1616 @@
...
+        ),
+        YieldFarmData<T, I>,
+        OptionQuery,
+    >;
+
+    /// Deposit details.
+    #[pallet::storage]
+    #[pallet::getter(fn deposit)]
+    pub type Deposit<T: Config<I>, I: 'static = ()> =
+        StorageMap<_, Blake2_128Concat, DepositId, DepositData<T, I>, OptionQuery>;
```
apopiak, line 291: Should be fine to use twox64concat for keys that the chain controls (which is the case for the DepositId here)

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1616 @@
...
+
+    #[pallet::call]
+    impl<T: Config<I>, I: 'static> Pallet<T, I> {}
+}
+
+impl<T: Config<I>, I: 'static> Pallet<T, I> {
+    /// Create new liquidity mining program with provided parameters.
+    ///
+    /// `owner` account have to have at least `total_rewards` balance. This funds will be
+    /// transferred from `owner` to farm account.
```
apopiak, line 326: sounds like you want to use `ensure_can_withdraw`?

apopiak (commented):
just some notes and nitpicks ^^

poliwop (commented):
The additional invariant test on the "pot" account looks correct.

liquidity-mining/Cargo.toml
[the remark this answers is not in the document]
enthusiastmartin: agreed

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1656 @@
...
+
+        /// Minimum number of periods to run liquidity mining program.
+        #[pallet::constant]
+        type MinPlannedYieldingPeriods: Get<Self::BlockNumber>;
+
+        /// The block number provider
+        type BlockNumberProvider: BlockNumberProvider<BlockNumber = Self::BlockNumber>;
+
+        /// Id used to identify amm pool in liquidity mining pallet.
+        type AmmPoolId: Parameter + Member + Clone + FullCodec;
```
enthusiastmartin, line 163: I believe at this point it is better to do the following as there are lot of clone calls for amm_pool_id.

and you can now remove all the "clone()" calls for amm_pool_id in the code.
```suggestion
        type AmmPoolId: Parameter + Member + Copy + FullCodec;
```

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1656 @@
...
+
+type AssetIdOf<T, I = ()> = <T as pallet::Config<I>>::CurrencyId;
+type PeriodOf<T> = <T as frame_system::Config>::BlockNumber;
+
+#[frame_support::pallet]
+pub mod pallet {
+    use super::*;
+
+    #[pallet::pallet]
+    #[pallet::without_storage_info]
```
enthusiastmartin, line 131: remove without_storage_info.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1656 @@
...
+use orml_traits::MultiCurrency;
+use scale_info::TypeInfo;
+use sp_arithmetic::{
+    traits::{CheckedDiv, CheckedSub},
+    FixedPointNumber, FixedU128, Perquintill,
+};
+use sp_std::convert::{From, Into, TryInto};
+
+type AssetIdOf<T, I = ()> = <T as pallet::Config<I>>::CurrencyId;
+type PeriodOf<T> = <T as frame_system::Config>::BlockNumber;
```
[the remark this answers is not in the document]
martinfridrich: This type represents "x: amount of blocks(Period), e.g in case of `updated_at: 10` it's 10-th period not a 10-th bock

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1656 @@
...
+        /// Invalid min. deposit was set for global farm.
+        InvalidMinDeposit,
+
+        /// Price adjustment multiplier can't be 0.
+        InvalidPriceAdjustment,
+    }
+
+    /// Id sequencer for `GlobalFarm` and `YieldFarm`.
+    #[pallet::storage]
+    #[pallet::getter(fn farm_id)]
```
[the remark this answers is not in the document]
martinfridrich: This returns last used id.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1656 @@
...
+        // Number of periods since last farm update.
+        let periods_since_last_update: Balance = TryInto::<u128>::try_into(
+            current_period
+                .checked_sub(&global_farm.updated_at)
+                .ok_or(ArithmeticError::Underflow)?,
+        )
+        .map_err(|_| ArithmeticError::Overflow)?;
+
+        let global_farm_account = Self::farm_account_id(global_farm.id)?;
+        let left_to_distribute = T::MultiCurrency::free_balance(global_farm.reward_currency, &global_farm_account);
```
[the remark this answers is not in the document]
martinfridrich: `free_balance` is only used for accounts controlled by this pallet, this accounts should never hold locked assets.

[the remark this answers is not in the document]
enthusiastmartin: i guess it is ok.

liquidity-mining/src/lib.rs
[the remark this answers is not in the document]
enthusiastmartin: last_farm_id then ?;-)

apopiak:
I assume this is the Basilisk integration?
https://github.com/galacticcouncil/Basilisk-node/pull/527

martinfridrich:
> I assume this is the Basilisk integration? [galacticcouncil/Basilisk-node#527](https://github.com/galacticcouncil/Basilisk-node/pull/527)

Yes, and there is also WIP implementation for stableswap liquidity minig but it's similar to this.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1678 @@
...
+                    //Delete yield farm.
+                    yield_farm.state = FarmState::Deleted;
+                    global_farm.decrease_live_yield_farm_count()?;
+
+                    //Cleanup if it's possible
+                    if yield_farm.can_be_flushed() {
+                        global_farm.decrease_total_yield_farm_count()?;
+
+                        *maybe_yield_farm = None;
+                    }
```
apopiak, line 848: why is it not necessary to do the same check (and possible remove) here for the global farm?

liquidity-mining/src/tests/deposit_lp_shares.rs
```diff
@@ -0,0 +1,577 @@
...
+
+        let deposited_amount = 25;
+        let yield_farm_id = GC_BSX_TKN2_YIELD_FARM_ID;
+        pretty_assertions::assert_eq!(
+            LiquidityMining::deposit_lp_shares(
+                global_farm_id,
+                yield_farm_id,
+                BSX_TKN2_AMM,
+                deposited_amount,
+                |_, _| { Ok(8_u128) }
```
apopiak, line 155: Why not use orml_tokens (or similar) here? You have it in your mock and would avoid hardcoding incorrect values here, no?

apopiak (commented):
ok, first review pass

liquidity-mining/src/tests/invariants.rs
```diff
@@ -0,0 +1,310 @@
...
+            let pot_balance_1 = Tokens::total_balance(REWARD_CURRENCY, &pot_account);
+            let yield_farm_balance_1 = Tokens::total_balance(REWARD_CURRENCY, &yield_farm_account);
+            let s_0 = pot_balance_0 + yield_farm_balance_0;
+            let s_1 = pot_balance_1 + yield_farm_balance_1;
+
+            assert_eq_approx!(
+                FixedU128::from((s_0, ONE)),
+                FixedU128::from((s_1, ONE)),
+                FixedU128::from((TOLERANCE, ONE)),
+                "invariant: global_farm_balance + yield_farm_balance"
```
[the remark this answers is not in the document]
martinfridrich: yes, this doesn't need tolerance.

liquidity-mining/src/tests/invariants.rs
```diff
@@ -0,0 +1,310 @@
...
+            let reward =
+                LiquidityMining::update_global_farm(&mut global_farm, current_period, reward_per_period).unwrap();
+
+            let s_0 = (left_to_distribute_0 - reward).max(0);
+            let s_1 = Tokens::free_balance(REWARD_CURRENCY, &global_farm_account);
+
+            assert_eq_approx!(
+                FixedU128::from((s_0, ONE)),
+                FixedU128::from((s_1, ONE)),
+                FixedU128::from((TOLERANCE, ONE)),
```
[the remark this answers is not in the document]
martinfridrich: You are right, in this case it doesn't needs to be fixed.

liquidity-mining/src/tests/full_run.rs
```diff
@@ -0,0 +1,606 @@
...
+        assert_ok!(LiquidityMining2::withdraw_lp_shares(
+            BOB_DEPOSIT,
+            YIELD_FARM_B,
+            unclaimable
+        ));
+
+        let (_, _, _, unclaimable) =
+            LiquidityMining2::claim_rewards(CHARLIE, CHARLIE_DEPOSIT, YIELD_FARM_B, false).unwrap();
+        assert_eq!(unclaimable, 0);
+        assert_ok!(LiquidityMining2::withdraw_lp_shares(CHARLIE, YIELD_FARM_B, unclaimable));
```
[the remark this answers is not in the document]
martinfridrich: yes, this is my mistake.
I distinguish between them so it's easier to see when it's account and when it's deposit. 
e.g if line above would be `LiquidityMining2::claim_rewards(CHARLIE, CHARLIE, ...)` it would be more confusing to know second `CHARLIE` is deposit id just by looking at it.

liquidity-mining/src/lib.rs
martinfridrich: Global farm can't be destroyed before all the yield farms are destroyed fist, so it can't be removed from storage.

https://github.com/galacticcouncil/warehouse/blob/fb4d7835598c009906e4b1bc0058373ea1b665b3/liquidity-mining/src/lib.rs#L469

Yield farm(active or stopped) can't exist in destroyed global farm. Destruction of global farm results in transfers of undistributed rewards to farm's owner so `active` yield farm can be no longer rewarded from global farm and `stopped` yield farm can be theoretically resumed. 
That's the reason why global farm can't have `active` or `stopped` yield farms before it's destroyed(marked for removal from storage).

liquidity-mining/src/tests/invariants.rs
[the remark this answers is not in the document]
apopiak: Then let's add the constant value to the invariant maybe?

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1678 @@
...
+    ///
+    /// Parameters:
+    /// - `who`: farm's owner
+    /// - `global_farm_id`: farm id to which a yield farm will be added.
+    /// - `multiplier`: yield farm multiplier.
+    /// - `loyalty_curve`: curve to calculate loyalty multiplier to distribute rewards to users
+    /// with time incentive. `None` means no loyalty multiplier.
+    /// - `amm_pool_id`: identifier of the AMM pool.
+    /// - `assets`: list of assets in the AMM pool. One of this assets must be incentivied asset
+    fn create_yield_farm(
```
apopiak, line 511: This should be `require_transactional` at the minimum because of the order of `YieldFarm::insert()` and `global_farm.increase_yield_farm_counts()?` (fallible function call after storage write)

liquidity-mining/src/tests/invariants.rs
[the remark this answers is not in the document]
apopiak: that was phrased confusingly, I mean lets add to the invariant text that it should be constant so e.g. "invariant: sum `global_farm_balance + yield_farm_balance` is always constant"

liquidity-mining/src/tests/full_run.rs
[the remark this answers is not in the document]
apopiak: Then let's just set `CHARLIE` (and the others) to a different number? (e.g. 33 or sth)

liquidity-mining/src/tests/claim_rewards.rs
```diff
@@ -0,0 +1,701 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+use test_ext::*;
+
+#[test]
+fn claim_rewards_should_work() {
```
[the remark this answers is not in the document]
apopiak: In general, the shorter (the code of) and more descriptive (the name and comments of) the test the better IMO

liquidity-mining/src/lib.rs
apopiak: This does not answer my question, I think. I'm aware of what you're describing.
I mean that it could be the case that we're removing the last yield farm of a global farm that already had `destroy_global_farm` called on it here and would thus not receive any further calls for this global farm that could do the final removal. Or would you expect the owner to call `destroy_global_farm` again? Seems superfluous to me if we can just remove it here.
I could probably write a test for this.

liquidity-mining/src/tests/claim_rewards.rs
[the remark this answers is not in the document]
martinfridrich: yes, I agree

liquidity-mining/src/lib.rs
martinfridrich: `destroy_global_farm` can't be called(it will fail) before all the yield farms are destroyed. Admin must fist destroyed all the yield farm first and as a last step `destroy_global_farm` can be called.

liquidity-mining/src/tests/deposit_lp_shares.rs
martinfridrich: We would not be able to use just `Tokens::free_balance(...)`, Account used by orml_tokens would still needs to be initialized with expected value.

liquidity-mining/src/lib.rs
apopiak: But `destroy_global_farm` can be called when all farms are set to `Deleted` but still in storage.

apopiak: I'll write a test I guess

apopiak: Ok, I'm having trouble coming up with a test, you might be right :sweat_smile:

apopiak: I thought that `destroy_yield_farm` could somehow be called after `destroy_global_farm`, but that's not the case.

liquidity-mining/src/tests/destroy_yield_farm.rs
```diff
@@ -0,0 +1,150 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+use test_ext::*;
+
+#[test]
+fn destory_yield_farm_with_deposits_should_work() {
```
apopiak, line 22: typo (also copy pasted in other parts of this file)

liquidity-mining/src/lib.rs
martinfridrich: yes, if I didn't make mistake, calling `destroy_yield_farm` after `destroy_global_farm` should not be possible. 
All the yield farms must be destroyed before destroying global farm.

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1678 @@
...
+
+            if !global_farm.total_shares_z.is_zero() && global_farm.updated_at != current_period {
+                let reward_per_period = math::calculate_global_farm_reward_per_period(
+                    global_farm.yield_per_period.into(),
+                    total_shares_z_adjusted,
+                    global_farm.max_reward_per_period,
+                )
+                .map_err(|_| ArithmeticError::Overflow)?;
+                Self::update_global_farm(global_farm, current_period, reward_per_period)?;
+            }
```
apopiak, line 439: Why is this part of adjusting the price? I think it makes more sense to separate the concerns

liquidity-mining/src/tests/claim_rewards.rs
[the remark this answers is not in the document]
martinfridrich: I created task for this

liquidity-mining/src/types.rs
```diff
@@ -0,0 +1,378 @@
...
+            accumulated_claimed_rewards: Zero::zero(),
+            entered_at,
+            updated_at: entered_at,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum FarmState {
```
apopiak, line 368: missing docs

liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1683 @@
...
+            yield_farm.accumulated_rpvs,
+            yield_farm.total_valued_shares,
+            yield_farm_rewards,
+        )
+        .map_err(|_| ArithmeticError::Overflow)?;
+        yield_farm.updated_at = current_period;
+
+        let pot_balance = T::MultiCurrency::free_balance(reward_currency, &Self::pot_account_id());
+
+        ensure!(pot_balance >= yield_farm_rewards, Error::<T, I>::InsufficientPotBalance);
```
apopiak, line 1371: missing test for this error

liquidity-mining/src/tests/tests.rs
```diff
@@ -0,0 +1,2186 @@
...
+                test_farm_entries[4].clone(),
+                test_farm_entries[6].clone(),
+            ]
+            .try_into()
+            .unwrap(),
+        }
+    );
+
+    //5 is max farm entries.
+    assert_err!(
```
apopiak, line 1829: should probably be `assert_noop`
```suggestion
    assert_noop!(
```
same for the other `assert_err`s

martinfridrich: `assert_err` is used when fns doesn't needs externalities. This are failable fns without storage operations.

apopiak: Ah, that makes sense, sorry I missed it.

apopiak (approved):
Looks good to me.
The complexity is quite high so I'm worried about having missed something, but I don't see a good way of addressing that.

PR #57, opened 2022-06-20
Collator Rewards without Session Keys Change

apopiak:


lumir-mrkva (commented):
do we still need `OneSessionHandler` implementation?

lumir-mrkva (approved):
tested it with runtime upgrade and it works like expected

apopiak:
> do we still need `OneSessionHandler` implementation?

Nope, part of why this was in draft is that it still contains the deprecated implementation.

collator-rewards/src/lib.rs
```diff
@@ -93,36 +95,42 @@ pub mod pallet {
...
-                        currency: T::RewardCurrencyId::get(),
-                    });
-                } else {
-                    log::warn!("Error reward collators: {:?}", result);
+impl<T: Config> SessionManager<T::AccountId> for Pallet<T> {
+    fn new_session(index: SessionIndex) -> Option<Vec<T::AccountId>> {
+        let maybe_collators = T::SessionManager::new_session(index);
+        if let Some(ref collators) = maybe_collators {
+            // stores the collators; will be removed in `end_session` to avoid storage piling up
+            Collators::<T>::insert(index, collators)
```
apopiak, line 110:
```suggestion
            Collators::<T>::insert(index, collators)
```

martinfridrich (approved):

PR #58, opened 2022-06-22
feat(nft): relocate permission checks

Roznovjak:


nft/src/lib.rs
```diff
@@ -126,6 +126,9 @@ pub mod pallet {
         ) -> DispatchResult {
             let sender = ensure_signed(origin)?;
 
+            ensure!(T::ReserveClassIdUpTo::get() < class_id, Error::<T>::IdReserved);
+            ensure!(T::Permissions::can_create(&class_type), Error::<T>::NotPermitted);
```
green-jay, line 130: @enthusiastmartin wanted to have the permission check (L130) inside of do_ functions so when implementing NFT traits the check is still there for types. We can move just ReserveClassIdUpTo (L129) check one level up.

Roznovjak: updated

dmoka (approved):
looks fine by me!

PR #61, opened 2022-07-05
add price for native currency to oracle

apopiak:


enthusiastmartin (changes requested):
Needs version bump but lgtm

PR #63, opened 2022-07-06
refactor: nft events, create functions

Roznovjak:


nft/src/lib.rs
```diff
@@ -309,15 +309,15 @@ impl<T: Config> Pallet<T> {
...
+        Ok(())
     }
 
     fn do_mint(
         owner: T::AccountId,
         class_id: T::NftClassId,
         instance_id: T::NftInstanceId,
         metadata: BoundedVecOfUnq<T>,
-    ) -> Result<T::NftInstanceId, DispatchError> {
+    ) -> DispatchResult {
```
lumir-mrkva, line 320: this is definetly not just test change, why did you sneak this in?

Roznovjak: Because it is a private method and the result is never used

green-jay (approved):

PR #66, opened 2022-07-11
Change Account Callback

apopiak:
...to not set fee currency if there is native balance in the account.


apopiak:
fixes issues in https://github.com/galacticcouncil/Basilisk-node/pull/514

lumir-mrkva (approved):

apopiak:
Note that session benchmarks will need to run again to update weights.

PR #69, opened 2022-07-19
feat:  CreateTypedClass and ReserveCassIdUpTo traits

martinfridrich:
* `CtrateTypedClass` trait to allow create typed classes from our pallets.
* `ReserveClassIdUpTo` trait for checking if id is from nft's reserved range. This is usable to check our pallets configuration e.g if we configured pallet's nft class from reserved range.

### DIFF
--- a/nft/src/lib.rs
+++ b/nft/src/lib.rs
@@ -29,6 +29,7 @@ use frame_support::{
 use frame_system::ensure_signed;
 use pallet_uniques::DestroyWitness;
 
+use hydradx_traits::nft::{CreateTypedClass, ReserveClassIdUpTo};
 use sp_runtime::{
     traits::{AtLeast32BitUnsigned, StaticLookup, Zero},
     DispatchError,
@@ -125,7 +126,7 @@ pub mod pallet {
         ) -> DispatchResult {
             let sender = ensure_signed(origin)?;
 
-            ensure!(T::ReserveClassIdUpTo::get() < class_id, Error::<T>::IdReserved);
+            ensure!(!Self::is_id_reserved(class_id), Error::<T>::IdReserved);
 
             Self::do_create_class(sender, class_id, class_type, metadata)?;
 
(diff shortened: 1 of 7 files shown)


nft/src/lib.rs
```diff
@@ -527,3 +528,15 @@ impl<T: Config> Transfer<T::AccountId> for Pallet<T> {
...
+
+impl<T: Config> CreateTypedClass<T::AccountId, T::NftClassId, T::ClassType> for Pallet<T> {
+    fn create_typed_class(owner: T::AccountId, class_id: T::NftClassId, class_type: T::ClassType) -> DispatchResult {
+        Self::do_create_class(owner, class_id, class_type, Default::default())
+    }
+}
+
+impl<T: Config> ReserveClassIdUpTo<T::NftClassId> for Pallet<T> {
+    fn is_id_reserved(id: T::NftClassId) -> bool {
+        id <= T::ReserveClassIdUpTo::get()
```
enthusiastmartin, line 540: have you considered specify the range instead of just max reserved id ?

i feel it would be more flexible, if you need to specify a range eg. 100..200 ?

traits/src/nft.rs
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+use frame_support::{dispatch::DispatchResult, traits::tokens::nonfungibles::Create};
+
+pub trait CreateTypedClass<AccountId, ClassId, ClassType>: Create<AccountId> {
+    /// This function reate nft class of `class_type` type.
+    fn create_typed_class(owner: AccountId, class_id: ClassId, class_type: ClassType) -> DispatchResult;
+}
+
+pub trait ReserveClassIdUpTo<ClassId> {
```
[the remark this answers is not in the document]
martinfridrich: I would prefer this `ReserveClassId::is_id_reserved(id:) -> bool` if we want to rename it but if you thing "validate" is better I can change it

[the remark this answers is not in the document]
enthusiastmartin: > I would prefer this `ReserveClassId::is_id_reserved(id:) -> bool` if we want to rename it but if you thing "validate" is better I can change it

can be. I mostly did not like the "UpTo" thing.

nft/src/lib.rs
enthusiastmartin: Fine with be. It was mainly a question whether there was a need to specify the minimum as well, different than 0.

enthusiastmartin (approved):

lumir-mrkva (approved):

traits/src/nft.rs
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use frame_support::{dispatch::DispatchResult, traits::tokens::nonfungibles::Create};
+
+pub trait CreateTypedClass<AccountId, ClassId, ClassType>: Create<AccountId> {
+    /// This function reate nft class of `class_type` type.
```
green-jay, line 21:
```suggestion
    /// This function creates an NFT class of `class_type` type.
```

traits/src/nft.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn create_typed_class(owner: AccountId, class_id: ClassId, class_type: ClassType) -> DispatchResult;
+}
+
+pub trait ReserveClassId<ClassId> {
+    /// This function return `true` if class id is from reserved range, `false` otherwise.
```
green-jay, line 26:
```suggestion
    /// This function returns `true` if class id is from the reserved range, `false` otherwise.
```

green-jay (approved):
lgtm

PR #71, opened 2022-07-22
fix: nft permissions

martinfridrich:
Changes:
Permissions checks was moved to extrinsics functions.

Reason:
With current implementation we can't use "internal" functions to create/mint/burn/... nfts from our pallets without allowing it in permissions which result in allowing action for everyone.


enthusiastmartin (commented):
looks ok to me. 

are there tests for this ?  I havent checked.

PR #72, opened 2022-07-27
fix: add missing impl for withdrawing tx fees

Roznovjak:


transaction-multi-payment/src/tests.rs
```diff
@@ -439,10 +440,14 @@ fn fee_payment_in_native_currency_work() {
             let tip = 0;
             let dispatch_info = info_from_weight(15);
 
+            // Act & Assert
```
dmoka, line 443: small thing, but if I am not mistaken, `//Act` starts at 443, and `//Assert` from 447, so I would split this comment, and place the //Assert correspondingly

Also true in some of the other tests.

Roznovjak: The reason why I placed both `Act` and `Assert` on the same line is that "act" is in these tests divided into two parts (pre_dispatch and post_dispatch) and each part is followed by some asserts. What do you think about using two `Act` and `Assert` sections in this case? Or do you prefer to use just one pair of `Act/Assert` on different lines as you suggested?

dmoka: Now I see, I overlooked `post_dispatch` in the lines of assertions.

Does `post_dispatch` have a dependency on `pre_dispatch`, or  `post_dispatch` can be called without `pre_dispatch` beforehand?

Because if they are unrelated, then I would split the tests into two test cases as right now we are testing two separate public APIs (behaviours) in one test.

If they are call-dependent, then I would do what you suggested, namely to place two `//Act`, one before `pre_dispatch` and one before `post_dispatch`, so we can easily figure out in the future which public APIs we are testing by just reading these comments. And of course, placing (two) //Assert before the corresponding assertion lines for each of them.

lumir-mrkva:
Task linked: [CU-2p79uye Upgrade to polkadot-v0.9.26](https://app.clickup.com/t/2p79uye)

transaction-multi-payment/src/tests.rs
Roznovjak: It's possible to test them separately, but the methods are coupled (the output of the first one is the input for the second one). It is also better to have them together because then it's easier to compare states before and after a transaction is executed.

dmoka (approved):

PR #76, opened 2022-08-05
feat: Next Generation Oracle

apopiak:
... using exponential moving averages (EMA).
See [EMA article](https://nestedsoftware.com/2018/04/04/exponential-moving-average-on-streaming-data-4hhl.24876.html) and [EMA Wikipedia section](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average).

New pallet that provides 4 oracles aggregated over different time periods (last block, 10 minutes, 1 day, 1 week).

Depends on https://github.com/galacticcouncil/HydraDX-math/pull/66

### Requirements
+ [x] oracles smoothing over different time periods (currently implemented: immediate, 10 minutes, 1 day, 1 week)
+ [x] price, volume and liquidity oracle
+ ~~[ ] return error if an oracle has not been active for long enough~~
+ [x] lazy oracles that update when receiving data `on_trade`/`on_liquidity_changed` or on `get_price`

### Open Questions
+ [x] ~~For which time period to return error for (until oracle is intialized)?~~
  + return age of oracle so consumers can decide for themselves
+ [x] Which time periods to offer?
+ [x] How to bound/weigh oracle updates (because of laziness)?
  + We use exponentiation so we don't pay the price for iteration and the weight does not increase with longer times between updates.
+ [x] What to do about switched order of assets? Adjust price and volume? Return predictable normalized version?
  + decided to normalize based on requested order
+ [x] Offer function for removing an oracle? (e.g. for XYK pool removal?)
  + not initially
+ [x] Move to new pallet instead of replacing the previous one?

### Todos
+ [x] Weights
+ [x] moar tests!
+ [x] extract math https://github.com/galacticcouncil/HydraDX-math/pull/66
+ [x] move to Polkadot 0.9.29 ~~0.9.28 0.9.26~~


price-oracle/src/lib.rs
```diff
@@ -126,6 +128,15 @@ pub mod pallet {
     #[pallet::getter(fn price_data_thousand)]
     pub type PriceDataThousand<T: Config> = StorageMap<_, Twox64Concat, AssetPairId, BucketQueue, ValueQuery>;
 
+    #[pallet::storage]
+    #[pallet::getter(fn accumulator)]
+    pub type Accumulator<T: Config> = StorageValue<_, BTreeMap<AssetPairId, PriceEntry>, ValueQuery>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn oracle)]
```
dmoka, line 136: maybe we can call this getter as `oracles` to follow the convention 
```suggestion
    #[pallet::getter(fn oracles)]
```

price-oracle/src/lib.rs
```diff
@@ -263,6 +285,28 @@ impl<T: Config> Pallet<T> {
                 });
             }
         }
+
+        // EMA oracles
+        for (pair_id, price_entry) in Accumulator::<T>::take().into_iter() {
+            Oracles::<T>::mutate(pair_id.clone(), 1, |oracle| {
```
dmoka, line 291: If possible, I would extract line 291-296, as the same logic is used 3 times in a row.

price-oracle/src/lib.rs
```diff
@@ -126,6 +128,15 @@ pub mod pallet {
...
     pub type PriceDataThousand<T: Config> = StorageMap<_, Twox64Concat, AssetPairId, BucketQueue, ValueQuery>;
 
+    #[pallet::storage]
+    #[pallet::getter(fn accumulator)]
+    pub type Accumulator<T: Config> = StorageValue<_, BTreeMap<AssetPairId, PriceEntry>, ValueQuery>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn oracle)]
+    pub type Oracles<T: Config> =
+        StorageDoubleMap<_, Twox64Concat, AssetPairId, Twox64Concat, u32, PriceEntry, OptionQuery>;
```
dmoka, line 138: I feel that by defining an alias for `key2` of type `u32`, we could make this storage more descriptive.

price-oracle/src/types.rs
```diff
@@ -102,6 +102,37 @@ impl PriceEntry {
...
+    }
+
+    pub fn calculate_new_ema_entry<const N: u32>(&self, previous_entry: &Self) -> Option<Self> {
+        use sp_arithmetic::{traits::One, FixedPointNumber};
+
+        let alpha = Price::saturating_from_rational(2u32, N.max(1) + 1);
+        debug_assert!(alpha <= Price::one());
+        let inv_alpha = Price::one() - alpha;
+
+        // TODO: include time
```
dmoka, line 122: todo :)

price-oracle/src/lib.rs
apopiak: The convention should be singular for maps (at least where it makes sense, might be different if you store a vec of things in a map entry) and has been at Parity IIRC. Using the getter I get one oracle entry, not several.

lumir-mrkva:
Task linked: [CU-2v0w820 next gen oracle](https://app.clickup.com/t/2v0w820)

price-oracle/src/lib.rs
dmoka: ah okay, I agree with you. I am working on a pallet right now (nft marketplace) and over there we have plural namings for both maps and getters, despite the fact that we store singular objects and not `vec` or so. But then I might need to do some refactorings on my pallet myself x)

Roznovjak (commented):
I would move `AssetPairId` type to types.rs . This type can also be changed to `BoundedVec<u8, 11>`. It might come handy later if we want to use bounded types for all the storage entries.

apopiak:
> I would move `AssetPairId` type to types.rs . This type can also be changed to `BoundedVec<u8, 11>`. It might come handy later if we want to use bounded types for all the storage entries.

I'm thinking about extracting it into an associated type, actually, and thus pulling it into the traits crate.

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,494 @@
...
+    let secs_per_block = T::SecsPerBlock::get();
+    let minutes = T::BlockNumber::from(60u8) / secs_per_block;
+    let days = T::BlockNumber::from(24u8) * T::BlockNumber::from(60u8) * minutes;
+    match period {
+        OraclePeriod::LastBlock => One::one(),
+        OraclePeriod::TenMinutes => T::BlockNumber::from(10u8) * minutes,
+        OraclePeriod::Day => days,
+        OraclePeriod::Week => T::BlockNumber::from(7u8) * days,
+    }
+}
```
apopiak, line 433: @poliwop Has suggested associating the smoothing factor directly with the oracle, skipping the path via block period.

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,494 @@
...
+    ///
+    /// Stores the data entry as well as the block number when the oracle was first initialized.
+    #[pallet::storage]
+    #[pallet::getter(fn oracle)]
+    pub type Oracles<T: Config> = StorageNMap<
+        _,
+        (
+            NMapKey<Twox64Concat, Source>,
+            NMapKey<Twox64Concat, (AssetId, AssetId)>,
+            NMapKey<Twox64Concat, T::BlockNumber>,
```
apopiak, line 118: If we drop the period I would replace this key with the Oracle identifier (I'm worried using FixedU128 as a key would get too messy.)
Another option if we don't want to change the set of oracles would be to collapse the different Oracles into one storage key:
```Rust
StorageDoubleMap<Twox64Concat, Source, Twox64Concat, (AssetId, AssetId), [(OracleEntry<T::BlockNumber>, T::BlockNumber); 4]>
```
Or similar

apopiak (commented):
Would be good to get your opinion on how to associate smoothing factors with oracle type.

ema-oracle/src/lib.rs
poliwop: not sure I fully understand the issue with using FixedU128 as a key... but if alpha (smoothing factor) is in (0,1], you could compute K=1/alpha as a u128 and use that as the key. I believe different alphas which have matching K would be so close we would likely not want to allow two different oracles with such close parameters anyway (assuming the division is done carefully to minimize rounding error).

For the purposes of the actual computation, I would still store the smoothing factor directly, but another option would be to just store K and recompute smoothing factor from it. However I wouldn't really see the advantage of that over the current implementation.

traits/src/oracle.rs
```diff
@@ -0,0 +1,272 @@
...
+            oracle_age,
+        }
+    }
+}
+
+/// Struct to represent trade volume for an asset pair.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(RuntimeDebug, Encode, Decode, Clone, PartialEq, Eq, Default, TypeInfo, MaxEncodedLen)]
+pub struct Volume<Balance> {
+    pub a_in: Balance,
```
dmoka, line 84: if possible, I would use more intention-revealing names here

traits/src/oracle.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub b_out: Balance,
+    pub a_out: Balance,
+    pub b_in: Balance,
```
dmoka, line 87: what is the reason having the fields in this order, namely ain-bout-aout-bin. Why not for example ain-aout-bin-bout?

traits/src/oracle.rs
```diff
@@ -0,0 +1,272 @@
...
+        Self {
+            a_in: Zero::zero(),
+            b_out: Zero::zero(),
+            a_out,
+            b_in,
+        }
+    }
+
+    /// Utility function that sums the underlying values of the volumes.
+    pub fn saturating_add(&self, rhs: &Self) -> Self {
```
dmoka, line 115: what does `rhs` mean? If it is a custom or less-known abbreviation, I would write it out.

traits/src/oracle.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    fn get_price_weight() -> Weight {
+        Weight::zero()
+    }
+}
+
+/// Mock implementation of the oracle trait that always returns `Price::one()` and oracle age of
+/// `BlockNumber::one()`.
+impl<AssetId, BlockNumber, Price> AggregatedPriceOracle<AssetId, BlockNumber, Price> for AlwaysPriceOfOne
```
dmoka, line 253: If this is only used in tests, then I would place it next to the tests.

ema-oracle/src/tests/invariants.rs
```diff
@@ -0,0 +1,66 @@
...
+    (any::<AssetId>(), any::<AssetId>()).prop_filter("asset ids should not be equal", |(a, b)| a != b)
+}
+
+fn non_zero_amount() -> impl Strategy<Value = Balance> {
+    any::<Balance>().prop_filter("balance should be greater 0", |b| b > &0)
+}
+
+// Tests
+proptest! {
+    #[test]
```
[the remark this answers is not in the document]
apopiak: the default is 256
I chose the config (or lack of config) to make sure the tests run quickly

traits/src/oracle.rs
apopiak: a_in and b_out mostly go together (you swap token a in and token b out), same for a_out and b_in

traits/src/oracle.rs
apopiak: it stands for right-hand-side and is a common abbreviation in math code

apopiak: I don't really care what it's called. We can change to `other`, too :shrug:

dmoka: ah I see it now. Just checked and there is also `rhs` used in other standard rust packages, f.e.: [u32](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add) 

Leave it then to follow the convention, and personally I learned something new, thanks for the info!

traits/src/oracle.rs
dmoka: ah okay, the ordering make sense then.

Just for my understanding: so a_in+b_out together represents a sell swap, and a_out+b_in represents a buy swap?

lumir-mrkva (changes requested):
pls merge master properly so the PR is not polluted with 0.9.28 related changes

traits/src/oracle.rs
apopiak: I don't think that's the right way to think about it? IIUC it doesn't matter for the oracle whether it was a buy or a sell, just whether token a moved into or out of the pool (and correspondingly for b).

traits/src/oracle.rs
apopiak: What direction are you thinking here?

traits/src/oracle.rs
apopiak: Do you mean in a test-util crate? Because this is meant for tests generally (e.g. also pallets that want to use the trait)

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+
+        /// Provider for the current block number.
+        type BlockNumberProvider: BlockNumberProvider<BlockNumber = Self::BlockNumber>;
+
+        /// The periods supported by the pallet. I.e. which oracles to track.
+        type SupportedPeriods: Get<BoundedVec<OraclePeriod, ConstU32<MAX_PERIODS>>>;
+
+        /// Maximum number of trades expected in one block.
+        #[pallet::constant]
+        type MaxTradesPerBlock: Get<u32>;
```
enthusiastmartin, line 94: i'd suggest to assert this is set to > 0 in integrity_test hook. The reason is that it is used in division later on ( not safe division ). Just to be sure.

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+            timestamp,
+        };
+        Pallet::<T>::on_trade(source, ordered_pair(asset_in, asset_out), entry);
+    }
+
+    fn on_trade_weight() -> Weight {
+        let max_trades = T::MaxTradesPerBlock::get();
+        // on_trade + on_finalize / max_trades
+        T::WeightInfo::on_trade_multiple_tokens(max_trades).saturating_add(
+            // TODO: Can we also divide the proof size weight the same way?
```
enthusiastmartin, line 338: TODO?

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+            timestamp,
+        };
+        Pallet::<T>::on_liquidity_changed(source, ordered_pair(asset_a, asset_b), entry);
+    }
+
+    fn on_liquidity_changed_weight() -> Weight {
+        let max_trades = T::MaxTradesPerBlock::get();
+        // on_liquidity + on_finalize / max_trades
+        T::WeightInfo::on_liquidity_changed_multiple_tokens(max_trades).saturating_add(
+            // TODO: Can we also divide the proof size weight the same way?
```
enthusiastmartin, line 380: TODO

ema-oracle/src/types.rs
```diff
@@ -0,0 +1,207 @@
...
+            price: self.price,
+            volume: self.volume,
+            liquidity: self.liquidity,
+            oracle_age: self.timestamp.saturating_sub(initialized),
+        }
+    }
+
+    /// Return an inverted version of the entry where the meaning of assets a and b are inverted.
+    /// So the price of a/b become the price b/a and the volume switches correspondingly.
+    pub fn inverted(&self) -> Self {
```
enthusiastmartin, line 68: not that important, although I'd think that you use the inverted entry and there would not be a need to keep the previous one and therefore you could consider:

```suggestion
    pub fn inverted(self) -> Self {
```

enthusiastmartin (approved):

ema-oracle/src/lib.rs
apopiak: I would create an issue, seems not worth addressing on the first pass, what do you think?

ema-oracle/src/types.rs
```diff
@@ -0,0 +1,207 @@
...
+        // It makes sense for the reciprocal of zero to be zero here.
+        let price = if self.price.is_zero() {
+            self.price
+        } else {
+            let (a, b) = self.price.into();
+            (b, a).into()
+        };
```
apopiak, line 75: @enthusiastmartin I'm actually not 100% sure what the semantics of an inverted price should be. I could also see someone arguing for max value or for the operation to fail. What do you think?

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub(crate) fn on_entry(src: Source, assets: (AssetId, AssetId), oracle_entry: OracleEntry<T::BlockNumber>) {
+        Accumulator::<T>::mutate(|accumulator| {
+            let is_present = if let Some(entry) = accumulator.get_mut(&(src, assets)) {
+                entry.accumulate_volume_and_update_from(&oracle_entry);
+                true
+            } else {
+                false
+            };
+            if !is_present {
+                accumulator.try_insert((src, assets), oracle_entry).unwrap_or_else(|((src, assets), entry)| {
```
apopiak, line 222: this should fail the trade -> change callback to result type

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,531 @@
...
+}
+
+impl<T: Config> Pallet<T> {
+    /// Insert or update data in the accumulator from received price entry. Aggregates volume and
+    /// takes the most recent data for the rest.
+    pub(crate) fn on_entry(src: Source, assets: (AssetId, AssetId), oracle_entry: OracleEntry<T::BlockNumber>) {
+        Accumulator::<T>::mutate(|accumulator| {
+            let is_present = if let Some(entry) = accumulator.get_mut(&(src, assets)) {
+                entry.accumulate_volume_and_update_from(&oracle_entry);
+                true
```
apopiak, line 217: return weight here

ema-oracle/src/lib.rs
apopiak: We should be able to divide proof size the same way as computational weight

traits/src/lib.rs
```diff
@@ -155,15 +157,29 @@ impl<AssetId> OnCreatePoolHandler<AssetId> for () {
...
 pub trait OnTradeHandler<AssetId, Balance> {
     /// Include a trade in the average price calculation of the price-oracle pallet.
-    fn on_trade(asset_a: AssetId, asset_b: AssetId, amount_in: Balance, amount_out: Balance, liq_amount: Balance);
+    fn on_trade(
+        source: Source,
+        asset_a: AssetId,
+        asset_b: AssetId,
+        amount_in: Balance,
+        amount_out: Balance,
+        liq_amount: Balance,
```
jak-pan, line 166: This will need to change to absolute amounts in the pool and as such it's probably mergable with the other handler but I'm not sure which one should we keep

jak-pan: We could call this on_before_trade or on_after_trade and sync up with volume limits

jak-pan: cc https://github.com/galacticcouncil/warehouse/pull/133/files#r1092011502

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,531 @@
...
+impl<T: Config> OnLiquidityChangedHandler<AssetId, Balance> for OnActivityHandler<T> {
+    fn on_liquidity_changed(
+        source: Source,
+        asset_a: AssetId,
+        asset_b: AssetId,
+        amount_a: Balance,
+        amount_b: Balance,
+        liquidity: Balance,
+    ) {
+        // We assume that zero values are not valid and can be ignored.
```
jak-pan, line 401: This could happen when the pool is destroyed

ema-oracle/src/lib.rs
```diff
@@ -0,0 +1,531 @@
...
+                    .saturating_sub(T::WeightInfo::on_finalize_no_entry())
+                    .ref_time()
+                    / (max_trades as u64),
+            ),
+        )
+    }
+}
+
+/// Calculate price from ordered assets
+pub fn determine_normalized_price(
```
jak-pan, line 435: This probably won't be the same for stableswap and omnipool. Do we need to make this more generic?

apopiak: This seems very unlikely to change. Happy to make it generic as soon as it needs to be.

ema-oracle/src/types.rs
apopiak: Ok, with the incoming amounts `on_liquidity_changed` being possibly zero (see https://github.com/galacticcouncil/warehouse/pull/76#discussion_r1092050361) the semantics should probably be that zero inverts to zero

Issue #52, opened 2022-06-03
refactor: registry traits


enthusiastmartin:
Registry traits should be refactored in a better way ( error generic as trait type,  types is interface consolidated etc ). 

Descriptions are needed too. 

_Originally posted by @Roznovjak in https://github.com/galacticcouncil/warehouse/pull/51#discussion_r888353502_

PR #82, opened 2022-09-19
feat: Stableswap v2

enthusiastmartin:


stableswap/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+            origin: OriginFor<T>,
+            assets: Vec<T::AssetId>,
+            amplification: u16,
+            trade_fee: Permill,
+            withdraw_fee: Permill,
+        ) -> DispatchResult {
+            T::CreatePoolOrigin::ensure_origin(origin)?;
+
+            let mut pool_assets = assets;
+            pool_assets.sort();
```
martinfridrich, line 281: I would suggest adding failing test with unsorted assets so this line is not removed by accident.

martinfridrich (commented):
I see potential way to merge iterations in `add_liquidity` into 1 `iter` but it would complicate inner code and also do more work before it fails so I'm not sure if it's worthed. Especially if set is that small. 
Also I see way how to get rid of `Btree`s. In my experience(from other lang) simple `find` is usually faster for small sets than construct more complicated data structure, but I didn't benchmarked it in rust.
If we are interested in this I can try to draft it.

enthusiastmartin:
> I see potential way to merge iterations in `add_liquidity` into 1 `iter` but it would complicate inner code and also do more work before it fails so I'm not sure if it's worthed. Especially if set is that small. Also I see way how to get rid of `Btree`s. In my experience(from other lang) simple `find` is usually faster for small sets than construct more complicated data structure, but I didn't benchmarked it in rust. If we are interested in this I can try to draft it.

i agree that it could be in one, however, this would make it quite complicated. 

because the thing is that add_liquidity can take any number of assets.  eg. if you have a pool with 5 assets, you can just add liquidity of one asset.
however you still need to load all asset's reserves and you also need to verify that all provided assets are in the pool. 

in this way, if there is an asset which does not belong to the pool - we might read balances of other assets only to find out that the input is incorrect.  

both approach have good and bad points.

stableswap/src/lib.rs
enthusiastmartin: This could be already checked by existing tests because it wont fail with unsorted assets.  the result might e different. so i believe existing tests would already fail, if it is not sorted. i will test.

martinfridrich: All the tests are passing if I comments this line. Tests are using sorted vec as input

stableswap/src/impls.rs
```diff
@@ -0,0 +1,71 @@
+use crate::{Balance, Config, Pallet, Zero};
```
dmoka, line 1: I would call this file `trade_execution.rs` to make it more descriptive.

We also use he same naming in LBP and XYK

stableswap/src/impls.rs
```diff
@@ -0,0 +1,71 @@
...
+use hydradx_traits::router::{ExecutorError, PoolType, TradeExecution};
+use sp_runtime::DispatchError;
+
+impl<T: Config> TradeExecution<T::Origin, T::AccountId, T::AssetId, Balance> for Pallet<T> {
```
dmoka, line 5: Since we add router support, we would also need to add support to Basilisk, right? I mean in the basilisk runtime route-executor config, amm pools associated type.

stableswap/src/lib.rs
```diff
@@ -0,0 +1,703 @@
...
+        LiquidityRemoved {
+            pool_id: T::AssetId,
+            who: T::AccountId,
+            shares: Balance,
+            asset: T::AssetId,
+            amount: Balance,
+            fee: Balance,
+        },
+        /// Sell trade executed. Trade fee paid in asset leaving the pool (already subtracted from amount_out).
+        SellExecuted {
```
dmoka, line 180: As `SellExecuted` and `BuyExecuted` has the exact same properties, shall we use just one `TradeExecuted` for both?

stableswap/src/lib.rs
```diff
@@ -0,0 +1,703 @@
...
+                T::AmplificationRange::get().contains(&amplification),
+                Error::<T>::InvalidAmplification
+            );
+            for asset in pool.assets.iter() {
+                ensure!(T::AssetRegistry::exists(*asset), Error::<T>::AssetNotRegistered);
+            }
+            let share_asset_ident = T::ShareAccountId::name(&pool.assets, Some(POOL_IDENTIFIER));
+            let share_asset = T::AssetRegistry::get_or_create_shared_asset(
+                share_asset_ident,
+                pool.assets.clone().into(), //TODO: fix the trait to accept refeerence instead
```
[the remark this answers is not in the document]
enthusiastmartin: removed todo.

should be refactor as part of https://github.com/galacticcouncil/warehouse/issues/52

stableswap/src/impls.rs
enthusiastmartin: I dont get what you mean. can you elaborate?

stableswap/src/lib.rs
enthusiastmartin: I'd keep it consistent with other pools where we have both sell/buy executed events.

stableswap/src/impls.rs
dmoka: what I mean is that right now we are in the warehouse project. In order to add the stableswap v2 to Basilisk or Hydra, we would need to extend the `pallet_route_executor` config in the given project.

So we would need to add Stableswap in `type AMM`. Or not?

image

enthusiastmartin: ah yes. when stableswap is added to basilisk, and support for the router is needed - then yes.

stableswap/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+[package]
+name = 'pallet-stableswap'
+version = '1.0.0'
+description = '2 asset stabelswap/curve AMM'
```
lumir-mrkva, line 4:
```suggestion
description = 'AMM for correlated assets'
```

stableswap/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        type Currency: MultiCurrency<Self::AccountId, CurrencyId = Self::AssetId, Balance = Balance>;
+
+        /// Account ID constructor
+        type ShareAccountId: AccountIdFor<Vec<Self::AssetId>, AccountId = Self::AccountId>;
+
+        /// Asset registry mechanism
+        type AssetRegistry: ShareTokenRegistry<Self::AssetId, Vec<u8>, Balance, DispatchError>;
+
+        /// The origin which can create a new pool
+        type CreatePoolOrigin: EnsureOrigin<Self::Origin>;
```
lumir-mrkva, line 122: this origin can also modify the existing pool. the more accurate name is required, for example: `PoolMasterOrigin`

lumir-mrkva (changes requested):

dmoka (approved):
LGTM.

Once stableswap is added to Basilisk, we could extend our router integration tests in Basilisk-node. It would be nice to have integration tests having 3 different kind of pools, so LBP, XYK and stableswap.

lumir-mrkva (approved):

PR #83, opened 2022-09-29
chore: upgrade to polkadot v0.9.28

Roznovjak:


adapters/src/lib.rs
```diff
@@ -36,15 +36,15 @@ mod tests;
 
 /// Weight trader that accepts multiple assets as weight fee payment.
 ///
-/// It uses `WeightToFee` in combination with a `NativePriceOracle` to set the right price for weight.
+/// It uses `ConvertWeightToFee` in combination with a `NativePriceOracle` to set the right price for weight.
```
dmoka, line 39: I think this one is still `WeightToFee`, as this is the name of the trait. For example after this you mention `NativePriceOracle` which is also a trait.

small thing...

```suggestion
/// It uses `WeightToFee` in combination with a `NativePriceOracle` to set the right price for weight.
```

currencies/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "pallet-currencies"
+description = "Provide `MultiCurrency` implementation using `pallet-balances` and `orml-tokens` module."
```
dmoka, line 3: I think it would be useful to add some info to the description about why we have this pallet custom recreated, and why we don't use the official one.

Or maybe, we could put this info in the README.

nft/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-nft"
-version = "3.3.1"
+version = "4.0.0"
```
dmoka, line 3: What is the main reason that we introduce a breaking change here?
 
I see that we need to do some migration, I guess it was necessary?

dmoka (approved):
Apart from some small things, looks fine to me, great job!

currencies/src/mock.rs
```diff
@@ -0,0 +1,179 @@
...
+type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Runtime>;
+type Block = frame_system::mocking::MockBlock<Runtime>;
+
+construct_runtime!(
+    pub enum Runtime where
+        Block = Block,
+        NodeBlock = Block,
+        UncheckedExtrinsic = UncheckedExtrinsic,
+    {
+        System: frame_system::{Pallet, Call, Storage, Config, Event<T>},
```
dmoka, line 118: as far as I know, these pallet definitions can be simplified in the latest version.

I would simplify them if possible.

route-executor/src/tests/mock.rs
```diff
@@ -52,7 +52,7 @@ frame_support::construct_runtime!(
          Router: router::{Pallet, Call,Event<T>},
```
dmoka, line 52: If possible, I would simplify these definitions.

nft/Cargo.toml
Roznovjak: It's a breaking change because the API, config, storage names, events and the errors has been changed thanks to renaming.

currencies/Cargo.toml
Roznovjak: I don't want to make any significant changes to this pallet because there is a high probability that if we update it to a newer version in the future, the code will be copy-pasted and any changes we made will be lost. I will add some comment to the runtimes that use the pallet.

nft/src/lib.rs
```diff
@@ -500,62 +505,70 @@ impl<T: Config> Create<T::AccountId> for Pallet<T> {
...
 
         Ok(())
     }
 
-    fn burn_from(class: &Self::ClassId, instance: &Self::InstanceId) -> DispatchResult {
-        let owner = Self::owner(*class, *instance).ok_or(Error::<T>::InstanceUnknown)?;
+    fn burn(
+        collection: &Self::CollectionId,
+        item: &Self::ItemId,
+        _maybe_check_owner: Option<&T::AccountId>,
```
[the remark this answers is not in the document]
Roznovjak: Thanks. Fixed.

PR #84, opened 2022-10-03
feat: add lbp type for router + refactor router to enforce individual trading limits

dmoka:
- LBP pool type has been added 
- router has been refactored to pass both amount in and out to trade execution, so we can use them as limits in the implementations


traits/src/router.rs
```diff
@@ -37,13 +38,15 @@ pub trait TradeExecution<Origin, AccountId, AssetId, Balance> {
         asset_in: AssetId,
         asset_out: AssetId,
         amount_in: Balance,
+        amount_out: Balance,
```
enthusiastmartin, line 41: why not name it precisely why it was added ? i believe the interface would be more clear. 

min limit ?

traits/src/router.rs
```diff
@@ -37,13 +38,15 @@ pub trait TradeExecution<Origin, AccountId, AssetId, Balance> {
...
     ) -> Result<(), ExecutorError<Self::Error>>;
 
     fn execute_buy(
         who: Origin,
         pool_type: PoolType<AssetId>,
         asset_in: AssetId,
         asset_out: AssetId,
+        amount_in: Balance,
```
enthusiastmartin, line 49: same here ? max_limit ? 

and i would then make as last param to be consistent.

lumir-mrkva (approved):

PR #85, opened 2022-10-14
feat: add missing tests

dmoka:
Added `test-utils` package where we can collect all the utility functions to be used in automated tests of pallets, such as asserting events, etc.

Did code coverage and mutation testing analysis, and based on that, added missing tests


adapters/src/tests.rs
```diff
@@ -393,4 +386,16 @@ fn revenue_goes_to_fee_receiver() {
     ExpectDeposit::register_expected_fee(42, CORE_ASSET_ID, 1234);
 
     Revenue::take_revenue((core_id, 1234).into());
+
+    assert_that_fee_is_deposited();
+}
+
+fn assert_that_fee_is_deposited() {
+    EXPECTED_DEPOSITS.with(|remaining| {
+        assert!(
```
apopiak, line 395: this has the unfortunate consequence that if this fails it will trigger for line 395 instead of 390. Not a big deal here, but annoying IMO if the function is reused
my personal workaround for cases like this is to define a new simple `macro_rules` macro, as it will generate the assert in the correct location

test-utils/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
+
+pretty_assertions = "1.3.0"
+
+
+[features]
+default = ["std"]
+
+std = [
+
```
apopiak, line 23:
```suggestion
  "frame-system/std",
```

test-utils/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+license = "Apache 2.0"
+repository = "https://github.com/galacticcouncil/warehouse"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
+
+pretty_assertions = "1.3.0"
```
apopiak, line 16: if this will be used in no-std contexts we should check whether pretty_assertions is no-std compatible

test-utils/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+use frame_system::Config;
+use pretty_assertions::assert_eq;
+
```
apopiak, line 3: we could add `assert_approx_eq!` to this as well

adapters/src/tests.rs
dmoka: that's a good one! One option could be to run the `cargo test` command with `RUST_BACKTRACE=1`, so we can see line 390 indicated. But as it requires additional effort, and is quite noisy, therefore I will go with a custom macro for that, like that!

test-utils/Cargo.toml
dmoka: I wonder why is this necessary. I mean, everything works properly without this line. What is the reason to add it in the std array?

test-utils/Cargo.toml
dmoka: This package will only be used for automated tests of pallets, at least that was my intention. As tests are in std contexts, I think we are fine with this?!

apopiak: Note that benchmarks are no_std, so you won't be able to use it there, then.

test-utils/Cargo.toml
apopiak: If this is an std package I would:
+ remove the `default-features = false` from the frame-system dep
+ remove the std feature (and the `default = [std]`)

test-utils/Cargo.toml
dmoka: I see! I think we are fine with having this as no_std, so only used in test code (excluding benchmark tests).

test-utils/Cargo.toml
dmoka: and if it is a `no_std` package?

apopiak: Then I would add the original above suggestion and make sure that `pretty_assertions` is used correctly (e.g make it optional if it depends on std)

test-utils/Cargo.toml
dmoka: Just figured out that the `pretty-assertions` has [no_std support](https://github.com/rust-pretty-assertions/rust-pretty-assertions#no_std-support), so we are good with this

test-utils/Cargo.toml
dmoka: All right, I will then add `frame-system/std` in `std` array. As we use pretty_assertions with no_std support, no need to make it optional, luckily.

lumir-mrkva (approved):

PR #92, opened 2022-10-31
chore: upgrade to polkadot-v0.9.29

enthusiastmartin:


transaction-multi-payment/src/lib.rs
```diff
@@ -87,7 +87,7 @@ pub mod pallet {
         fn on_initialize(_n: T::BlockNumber) -> Weight {
             let native_asset = T::NativeAssetId::get();
 
-            let mut weight: Weight = 0;
+            let mut weight: u64 = 0;
```
Roznovjak, line 90: We can use `Weight` struct here

Roznovjak (approved):

lumir-mrkva (approved):

PR #105, opened 2022-11-08
feat(nft): enable maybe_check_owner in nft destroy trait

### LINKED ISSUE #94: Ownership checking is not enforced in NFT interfaces
https://github.com/galacticcouncil/warehouse/blob/89bde6d2b8bad2263c554eb3d83b76d877971136/nft/src/lib.rs#L517
https://github.com/galacticcouncil/warehouse/blob/89bde6d2b8bad2263c554eb3d83b76d877971136/nft/src/lib.rs#L542

In the above-quoted code, `_maybe_owner_check` is passed into the function by the caller, application pallet. However, this argument is not processed in these interface functions. This has created confusion between the assumption and the implementation. 

The assumption could implicitly be that `the caller function would do the ownership checking`. However, such an assumption is not found in any documentation. 
The implementation (with the argument `_maybe_check_owner`) suggests that the ownership checking would be performed by the interface function where the caller function would configure whether ownership checking is needed using the argument.

Thus the boundary between the caller and the interface functions are not clear. It is however recommended to be consistent with the current interface design and implement the ownership checking by 
- if `_maybe_check_owner` matches `Some(check_owner)`, then `ensure!(check_owner == owner, Error::<T>::NotPermitted) 
- Otherwise, continue to call `Self::do_destroy_collection(owner, collection)?;`

Similar case for `burn()` interface.

Roznovjak:
Fixes #94

### DIFF
--- a/nft/src/lib.rs
+++ b/nft/src/lib.rs
@@ -511,7 +511,7 @@ impl<T: Config> Destroy<T::AccountId> for Pallet<T> {
         let owner = if let Some(check_owner) = maybe_check_owner {
             check_owner
         } else {
-            Self::collection_owner(collection).ok_or(Error::<T>::CollectionUnknown)?
+            Self::collection_owner(&collection).ok_or(Error::<T>::CollectionUnknown)?
         };
 
         Self::do_destroy_collection(owner, collection)?;
--- a/nft/src/lib.rs
+++ b/nft/src/lib.rs
@@ -514,9 +514,13 @@ impl<T: Config> Destroy<T::AccountId> for Pallet<T> {
     fn destroy(
         collection: Self::CollectionId,
         _witness: Self::DestroyWitness,
-        _maybe_check_owner: Option<T::AccountId>,
+        maybe_check_owner: Option<T::AccountId>,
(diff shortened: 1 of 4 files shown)


nft/src/lib.rs
```diff
@@ -514,9 +514,13 @@ impl<T: Config> Destroy<T::AccountId> for Pallet<T> {
     fn destroy(
         collection: Self::CollectionId,
         _witness: Self::DestroyWitness,
-        _maybe_check_owner: Option<T::AccountId>,
+        maybe_check_owner: Option<T::AccountId>,
```
dmoka, line 517: why is it called `maybe_check_owner`? It is a bit confusing, I mean, nothing is checked, as far as I can see.

maybe `maybe_collection_owner`?

dmoka (approved):

nft/src/lib.rs
Roznovjak: That's how it's named in the `Destroy` trait. And it's passed to pallet_uniques where it's checked. I agree with you that it's confusing. It's because we are implementing some other traits created by someone else.

PR #112, opened 2022-11-10
test(nft): add integration test for zero `ItemDeposit`

### LINKED ISSUE #90: Arbitrary minting could become an attack if `ItemDeposit` is not 0
The NFT pallet calls to [substrate::frame::uniques](https://github.com/paritytech/substrate/blob/0246883c404d498090f33e795feb8075fa8d3b6b/frame/uniques/src/functions.rs#L166) pallet to perform operations towards NFTs.
In terms of minting (and create_collection is similar), the design logic of the unique pallet would require [reserving](https://github.com/paritytech/substrate/blob/0246883c404d498090f33e795feb8075fa8d3b6b/frame/uniques/src/functions.rs#L166)(code linked) a certain amount from the collection owner's free balances.

The intended design in the NFT pallet however allows anyone to mint an item in this collection if the collection type has the minting permission. This would lead to a possible attack that anyone can mint NFT items using the collection owner's money causing the owner's fund to be locked if the item can not be sold.

The above observation is communicated to the development team. The current approach is to set `ItemDeposit` as 0 all the time to avoid the potential hazard. The same should be done with `CollectionDeposit` 

It is then strongly recommended to enforce condition checking or test cases against `runtime` to prevent accidental errors making these constants non-zero. Up-to-date documentation of this assumption is a must too.

Roznovjak:
Fixes #90


enthusiastmartin:
no longer relevant

PR #130, opened 2022-11-28
fix: LM global farm's paid accumulated rewards

### LINKED ISSUE #115: `paid_accumulated_rewards` not updated when rewards are returned by yield farm
When a yield farm is destroyed and it returns unused rewards, these rewards should be subtracted from the  `paid_accumulated_rewards` field of the corresponding global farm. Otherwise, such returned rewards can contribute several times to `paid_accumulated_rewards`, which can then exceed `total_rewards`.

martinfridrich:
This PR fixes #115 and refactoring `Deleted` state of farm to `Terminated`

### DIFF
--- a/liquidity-mining/Cargo.toml
+++ b/liquidity-mining/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "1.1.1"
+version = "1.1.2"
 description = "Liquidity mining"
 authors = ["GalacticCouncil"]
 edition = "2021"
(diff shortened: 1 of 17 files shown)


liquidity-mining/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "1.1.0"
+version = "1.1.2"
```
lumir-mrkva, line 3: change in API should trigger major version change

lumir-mrkva (approved):

PR #133, opened 2022-12-01
feat: add net trade volume limit per block

dmoka:
newer revision in https://github.com/galacticcouncil/HydraDX-node/pull/483


traits/src/lib.rs
```diff
@@ -143,33 +137,41 @@ pub trait OnCreatePoolHandler<AssetId> {
...
     /// Add this weight to an extrinsic from which you call `on_trade`.
-    fn on_trade_weight() -> Weight;
-}
-
-impl<AssetId, Balance> OnTradeHandler<AssetId, Balance> for () {
-    fn on_trade(_asset_a: AssetId, _asset_b: AssetId, _amount_in: Balance, _amount_out: Balance, _liq_amount: Balance) {
-    }
     fn on_trade_weight() -> Weight {
         Weight::zero()
     }
```
enthusiastmartin, line 165: why the need to change this and have the some default implemenations ? 

why you include same methods in this trait , eg "before_pool_state_change" ?

traits/src/lib.rs
```diff
@@ -143,33 +137,41 @@ pub trait OnCreatePoolHandler<AssetId> {
...
 }
 
+impl<AssetId, Balance> OnTradeHandler<AssetId, Balance> for () {}
```
enthusiastmartin, line 168: why the need to change ?

traits/src/lib.rs
```diff
@@ -143,33 +137,41 @@ pub trait OnCreatePoolHandler<AssetId> {
...
+
+pub trait OnPoolStateChangeHandler<AssetId, Balance> {
+    fn before_pool_state_change(asset_id: AssetId, initial_liquidity: Balance) -> DispatchResult;
```
enthusiastmartin, line 171: i'd prefer to name it someting other than "initial_liquidity" . Initial liquidity usually refers to liquidity at time of pool creation. 

Here is just current liquidity, or current reserve of the asset in the pool. 

you can just name it asset_reserve.

traits/src/lib.rs
Roznovjak: forgot to remove it. We have decided to create a new trait.

traits/src/lib.rs
Roznovjak: ... reverting it back

traits/src/lib.rs
Roznovjak: renamed

traits/src/lib.rs
```diff
@@ -170,6 +164,46 @@ impl<AssetId, Balance> OnTradeHandler<AssetId, Balance> for () {
     }
 }
 
+pub trait OnPoolStateChangeHandler<AssetId, Balance> {
+    fn before_pool_state_change(asset_id: AssetId, asset_reserve: Balance) -> DispatchResult;
+    fn after_pool_state_change(
```
jak-pan, line 169: https://github.com/galacticcouncil/warehouse/pull/76/files#r1092007628 these handlers can be merged after the change in oracles. Can you pls sync on this? I would also call this liquidity_out it it's liquidity_out and amount_out if it's amount_out (not sure which one you use here)

traits/src/lib.rs
```diff
@@ -170,6 +164,46 @@ impl<AssetId, Balance> OnTradeHandler<AssetId, Balance> for () {
...
+    ) -> DispatchResult {
+        Ok(())
+    }
+}
+
+/// Handler used by AMM pools to perform some tasks when liquidity changes outside of trades.
+pub trait OnLiquidityChangeHandler<AssetId, Balance> {
+    /// Notify that the liquidity for a pair of asset has changed.
+    fn before_add_liquidity(asset_id: AssetId, amount: Balance) -> DispatchResult;
+    fn after_add_liquidity(asset_id: AssetId, amount: Balance) -> DispatchResult;
```
jak-pan, line 195: Why do you need both of these? Isn't the one after enough (since it passed checks) + the amount that was added?

traits/src/lib.rs
jak-pan: I think you could use just one handler here but I didn't see impl yet might scratch this later.

traits/src/lib.rs
apopiak: I suspect they snuck in when merging :sweat_smile:

Roznovjak: Hmm. You are right. It's probably left over from the previous implementations. Just one function is enough. And the same also applies to `OnPoolStateChangeHandler`.

jak-pan:
Was this merged in different PR?

dmoka:
> Was this merged in different PR?

It was not merged, and was not needed. Eventually, we want to extract the trade/liquidty handling logic from Hydra to this warehouse repo so it can be also used in basilisk. As we were in hurry last week, and it would have not added much value, therefore we did not do that, for now

Issue #121, opened 2022-11-17
Suggestions towards code improvement


yanliu18:
Here is a list of suggestions to improve the code to avoid confusion and achieve better maintainability.

- [x]  Avoid using magic numbers, for example:
https://github.com/galacticcouncil/warehouse/blob/26afb15117a47314ef2d5d13730aeadf80d3609d/liquidity-mining/src/lib.rs#L1426
1 to which `min_deposit` is comparing, should be a constant variable rather than a magic number. This could improve code readability and maintainability.

- [x] Avoid confusion between `farm status` and `storage status`
https://github.com/galacticcouncil/warehouse/blob/174065dab263f45cdedbd528ce85a37b4065ee01/liquidity-mining/src/types.rs#L364-L368

It is recommended to use `Suspended` and `Terminated` to replace `Stopped` and `Deleted` respectively. In such a way, it would be clear that `deleted` or `destroyed` are referring to the farm being `removed` from storage. 

- [x] This part of the code is redundant, there will be an `ensure_can_withdraw()` checking in the `transfer` function
https://github.com/galacticcouncil/warehouse/blob/174065dab263f45cdedbd528ce85a37b4065ee01/liquidity-mining/src/lib.rs#L1390-L1392
If for the purpose of better error information, it is recommended to capture the error event from `transfer` function. Provided that it is assumed the `pot` and the `yield_farm_account` should always be alive, the error return by transfer should only be `LiquidityRestriction` which could be further mapped to `InsufficientPotBalance`.

PR #136, opened 2022-12-06
feat: limit global-farm's `min_deposit` and yield-farm's `multiplier`

### LINKED ISSUE #127: Important invariants possibly broken by small deposits due to rounding.
Because a yield farm's stake in the global farm is computed using the farm's `multiplier`, which can be `<1` (indeed, it is only asserted to be `>0`), it is possible that a non-zero deposit results in a `0` stake (or in other words, `0` `shares_z) in the global farm. This means a global farm can (incorrectly)  appear to be empty while hosting yield farms that (correctly) appear to be non-empty. 

Status: The current suggestion by the HydraDx team is to enforce a stricter lower-bound on both number of shares in a deposit and the value of `multiplier`. To be analyzed further.

martinfridrich:
This PR fixes:
*  #127 
* #121 (paratially)
and updates rust-toolchain.


liquidity-mining/src/tests/claim_rewards.rs
```diff
@@ -580,11 +579,14 @@ fn claim_rewards_doubleclaim_in_the_same_period_should_not_work() {
                 LiquidityMining::yield_farm((BSX_TKN1_AMM, GC_FARM, GC_BSX_TKN1_YIELD_FARM_ID))
                     .unwrap()
                     .left_to_distribute,
-                yield_farm_0.left_to_distribute - 23_306
+                yield_farm_0.left_to_distribute - 23_306_074_766_355_140
             );
 
-            pretty_assertions::assert_eq!(Tokens::free_balance(BSX, &ALICE), alice_bsx_balance + 23_306);
-            pretty_assertions::assert_eq!(Tokens::free_balance(BSX, &pot), pot_balance_0 - 23_306);
+            pretty_assertions::assert_eq!(
```
dmoka, line 585: Usually I just import `assert_eq` at the top of the file with `use pretty_assertions::assert_eq;`, then no need to specify it everywhere. Small detail...

liquidity-mining/src/tests/mod.rs
```diff
@@ -221,7 +221,7 @@ fn is_approx_eq_fixedu128(num_1: FixedU128, num_2: FixedU128, delta: FixedU128)
     };
 
     if diff.cmp(&delta) == Ordering::Greater {
-        println!("diff: {:?}; delta: {:?}; n1: {:?}; n2: {:?}", diff, delta, num_1, num_2);
+        println!("diff: {diff:?}; delta: {delta:?}; n1: {num_1:?}; n2: {num_2:?}");
```
dmoka, line 224: is this a new syntax because we updated the toolchain? Never seen it, like it!

liquidity-mining/src/lib.rs
```diff
@@ -557,7 +565,10 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
         amm_pool_id: T::AmmPoolId,
         assets: Vec<T::AssetId>,
     ) -> Result<YieldFarmId, DispatchError> {
-        ensure!(!multiplier.is_zero(), Error::<T, I>::InvalidMultiplier);
+        ensure!(
```
dmoka, line 568: in the corresponding test `add_yield_farm_invalid_multiplier_should_not_work`, I would also check the edge case so `FixedU128::from_inner(1_000_000_000_000_000 - 1)`, and not only `FixedU128::from(0_u128)`. You could utilize the [test-crate](https://docs.rs/test-case/latest/test_case/) to avoid test duplication.

dmoka (approved):
Apart from some small things, LGTM.

asset-registry/src/lib.rs
```diff
@@ -456,12 +456,12 @@ impl<T: Config> Pallet<T> {
...
 impl<T: Config> Registry<T::AssetId, Vec<u8>, T::Balance, DispatchError> for Pallet<T> {
     fn exists(asset_id: T::AssetId) -> bool {
-        Assets::<T>::contains_key(&asset_id)
+        Assets::<T>::contains_key(asset_id)
     }
 
     fn retrieve_asset(name: &Vec<u8>) -> Result<T::AssetId, DispatchError> {
         let bounded_name = Self::to_bounded_name(name.clone())?;
-        if let Some(asset_id) = AssetIds::<T>::get(&bounded_name) {
+        if let Some(asset_id) = AssetIds::<T>::get(bounded_name) {
```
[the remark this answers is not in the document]
martinfridrich: Probably yes, but refactor of asset-registry is not the purpose of this PR. I touched it only to make clippy happy

liquidity-mining/src/tests/mod.rs
martinfridrich: It was added in 1.65 and it was suggested by clippy

liquidity-mining/src/tests/create_yield_farm.rs
```diff
@@ -295,7 +296,7 @@ fn add_yield_farm_invalid_multiplier_should_not_work() {
                 LiquidityMining::create_yield_farm(
                     ALICE,
                     ALICE_FARM,
-                    FixedU128::from(0_u128),
+                    crate::MIN_YIELD_FARM_MULTIPLIER - FixedU128::from_inner(1),
```
dmoka, line 299: the issue with using the `MIN_YIELD_FARM_MULTIPLIER` constant here is that if the constant is changed by someone, let's say accidentally, then this test won't fail. And I think in such cases we want to have a failing test which can raise awareness and provide safety net.

So I would use specific values.

martinfridrich: hm, I'll change it but it can lead to the same situation as now. 
I did set `MIN_YIELD_FARM_MULTIPLIER` to higher value and fixed all the broken tests but forgot about this one because it's was  passing with `0` .

Basically this is the only important rule/test for `MIN_YIELD_FARM_MULTIPLIER` and `MIN_DEPOSIT`: https://github.com/galacticcouncil/warehouse/blob/9fbfcac5baf4ea6a4af9efcb9586687043f0d997/liquidity-mining/src/tests/tests.rs#L2195

PR #138, opened 2022-12-12
feat: add balance check utility methods

dmoka:


test-utils/src/lib.rs
```diff
@@ -32,3 +32,29 @@ macro_rules! assert_eq_approx {
...
+
+#[macro_export]
+macro_rules! assert_balance {
+    ($who:expr, $asset_id:expr,  $expected_balance:expr) => {{
+        assert_eq!(Tokens::free_balance($asset_id, &$who), $expected_balance);
+    }};
+}
+
+#[macro_export]
+macro_rules! assert_balance_approx {
```
enthusiastmartin, line 44: why would you need aprox balance ? balance of account should really be exact ?

lumir-mrkva (approved):

test-utils/src/lib.rs
dmoka: I think it would be just only used in invariant tests.

PR #139, opened 2022-12-14
fix: loyalty multiplier calculation in resumed yield-farm

martinfridrich:
This PR fixes: 
 * loyalty factor calculation when yield-farm is resumed
 * bug in rewards allocation from the global-farm after yield-farm resume if yield-farm is the only farm in the global farm.

# Loyalty factor fix:
## Problem:
Loyalty-multiplier is calculated based on the number of passed periods. Current period is calculated based on the current block number so it continues to grow. If yield-farm is stopped, passed periods are calculated based on yield-farm’s `updated_at` which results in correct loyalty-multiplier calculation only for non-stopped periods. 
If yield-farm is resumed, passed periods are calculated again based on current period and this results in growth of loyalty-multiplier for stopped periods.

## Fix:
https://github.com/galacticcouncil/warehouse/blob/502e5d469359d9150fe9d537c3a62097a6f4c986/liquidity-mining/src/lib.rs#L1144-L1157

# Rewards allocation fix:
## Problem:
If a global-farm has only 1 yield-farm and the yield-farm is stopped and resumed, global-farm’s RPZ update which should happen at yield-farm resume won’t happen because global-farm has zero `sharesZ`. This results in allocation of rewards from global-farm for stopped periods  on next global-farm’s RPZ update.

## Fix:
https://github.com/galacticcouncil/warehouse/blob/68d0018535818823a71ef2d90897d738922de35c/liquidity-mining/src/lib.rs#L892-L896


liquidity-mining/src/lib.rs
```diff
@@ -881,8 +881,22 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+                        .checked_sub(&yield_farm.updated_at)
+                        .defensive_ok_or::<Error<T, I>>(InconsistentStateError::InvalidPeriod.into())?;
+
                     Self::maybe_update_global_farm_rpz(global_farm, current_period)?;
 
+                    //NOTE: this is special case. Without this if global-farm has only 1 stopped
+                    //yield-farm and resume_yield_farm() is called, global-farm's update won't happen because it
+                    //has zero shares Z and next global-farm update will calulate rewards for
+                    //"empty" periods.
+                    if global_farm.updated_at != current_period {
```
enthusiastmartin, line 896: i think you can easily remove the condition check. if it already equal, the assignment wont make difference

liquidity-mining/src/types.rs
```diff
@@ -191,6 +192,7 @@ impl<T: Config<I>, I: 'static> YieldFarmData<T, I> {
             state: FarmState::Active,
             entries_count: Default::default(),
             left_to_distribute: Default::default(),
+            total_stopped: Default::default(),
             _phantom: PhantomData::default(),
```
enthusiastmartin, line 196: just wondering why phantom here is needed at all.

enthusiastmartin (approved):
from the code perspective is fine. 

however, it is very hard to verify the logic, as the feature is complex and requires familiarity.

liquidity-mining/src/lib.rs
martinfridrich: yes, true.

liquidity-mining/src/types.rs
martinfridrich: `I` is not used without  `phantomData<I>` and rust doesn't like it.

PR #150, opened 2023-01-17
extend asset registry genesis config to allow specifying ids at genesis

apopiak:
This is useful for e.g. integration tests where we might want to set e.g. `LRNA` at genesis to have a certain asset id.


apopiak:
Do we follow semver here? I would need to bump to 2 because it's part of the interface, yes?

enthusiastmartin:
> Do we follow semver here? I would need to bump to 2 because it's part of the interface, yes?

yes.

PR #154, opened 2023-01-25
feat: return amm_pool_id from redeposit_lp_shares

martinfridrich:
This PR updates liquidity-mining's `redeposit_lp_shares()` function to return also `amm_pool_id`. This change is necessary for Basilisk's xyk-liqudity-mining pallet to correctly validate `AssetPair` param from user.


Roznovjak (commented):
Please also update the docs, line 1039.

Roznovjak (approved):

PR #158, opened 2023-01-27
refactor: move substrate, cumulus and orml deps to workspace

apopiak:


enthusiastmartin:
I guess there is something wrong with scale codec. perhaps because of the crate name change in each cargo file. 

question: does this work correctly too where a single crate is used in another project ? i believe it should, just wondering.

apopiak:
> I guess there is something wrong with scale codec. perhaps because of the crate name change in each cargo file.

yeah, I've been working locally on fixing that

> question: does this work correctly too where a single crate is used in another project ? i believe it should, just wondering.

you mean someone depending on a warehouse pallet? I would hope so, might be worth testing

apopiak:
this is stale, will retry again later

PR #159, opened 2023-01-27
feat: expect unordered events

Roznovjak:


test-utils/src/lib.rs
```diff
@@ -9,6 +11,17 @@ where
...
 
+/// Test that the events were emitted, in no particular order.
+/// Provided events don't need to be the last events that were emitted.
+pub fn expect_unordered_events<TEvent: std::fmt::Debug + PartialEq, TRuntime: Config>(e: Vec<TEvent>)
+where
+    <TRuntime as Config>::Event: From<TEvent>,
+{
+    e.into_iter()
+        .map(|x| x.into())
+        .for_each(frame_system::Pallet::<TRuntime>::assert_has_event);
```
dmoka, line 22: I wonder if we could do this in a better way. The problem is that if `assert_has_event` fails then the error message is really cryptic, with not much value about the difference.

dmoka (approved):

test-utils/src/lib.rs
dmoka: maybe we could check how the `assert_has_event` is implemented, and doing our own implementation with pretty-assertions?

Roznovjak:
closing it because we didn't get the expected result.

PR #164, opened 2023-02-09
refactor(registry): genesis config

Roznovjak:


asset-registry/src/lib.rs
```diff
@@ -159,10 +159,10 @@ pub mod pallet {
     pub type AssetMetadataMap<T: Config> =
         StorageMap<_, Twox64Concat, T::AssetId, AssetMetadata<BoundedVec<u8, T::StringLimit>>, OptionQuery>;
 
+    #[allow(clippy::type_complexity)]
```
enthusiastmartin, line 162: about what the clippy complains ?

enthusiastmartin (approved):

asset-registry/src/lib.rs
Roznovjak: changing `AssetId` to `Option<AssetId>`

PR #165, opened 2023-02-11
feat: OTC desk

vgantchev:
Related PR:
https://github.com/galacticcouncil/HydraDX-node/pull/515

This PR introduces a pallet-otc which enables over-the-counter trading.


Cargo.toml
```diff
@@ -16,7 +16,8 @@ members = [
 	"liquidity-mining",
 	"currencies",
 	"stableswap",
-	"test-utils"
+	"test-utils",
+	"otc"
```
enthusiastmartin, line 20:
```suggestion
	"otc",
```

liquidity-mining/Cargo.toml
```diff
@@ -21,8 +21,6 @@ orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-li
 
 # HydraDX dependencies
 hydradx-traits = { path = "../traits", default-features = false }
-#NOTE: This is bacause we are not using latest version of asset-registry for now.
-registry-traits = { package="hydradx-traits", git = "https://github.com/galacticcouncil/warehouse", rev="fcebd285141ce9ed15136f6102fa6396605d96af", default-features = false }
```
enthusiastmartin, line 25: why this is resolved in the PR

liquidity-mining/src/lib.rs
```diff
@@ -115,9 +115,8 @@ use frame_system::pallet_prelude::BlockNumberFor;
 use sp_runtime::ArithmeticError;
 
 use hydra_dx_math::liquidity_mining as math;
-use hydradx_traits::pools::DustRemovalAccountWhitelist;
+use hydradx_traits::{pools::DustRemovalAccountWhitelist, registry::Registry};
```
enthusiastmartin, line 118: why any changes in lm in this PR

otc/README.md
```diff
@@ -0,0 +1 @@
+TODO
```
enthusiastmartin, line 1: todo

otc/Cargo.toml
```diff
@@ -0,0 +1,77 @@
...
+frame-support = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.29", default-features = false }
+frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.29", default-features = false }
+pallet-balances = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.29", default-features = false, optional = true }
+
+# Math
+hydra-dx-math = { git = "https://github.com/galacticcouncil/HydraDX-math", rev="2d7f86ffae242fc855bc833dd22df2d8dc5d03df", default-features = false }
+
+# HydraDX dependencies
+hydradx-traits = { path = "../traits", default-features = false }
+pallet-currencies = { path = "../currencies", default-features = false }
```
enthusiastmartin, line 35: it seems that tsome of the depedencies are not needed in the actual implemeantion:

- balances
- math
- traits
- currencies

maybe others ?

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+            + MaxEncodedLen
+            + TypeInfo;
+
+        /// Asset Registry mechanism - used to check if asset is correctly registered in asset registry
+        type AssetRegistry: Registry<Self::AssetId, Vec<u8>, Balance, DispatchError>;
+
+        /// Named reservable multi currency
+        type Currency: MultiCurrency<Self::AccountId, CurrencyId = Self::AssetId, Balance = Balance>
+            + MultiCurrencyExtended<Self::AccountId>
+            + NamedMultiReservableCurrency<Self::AccountId, ReserveIdentifier = NamedReserveIdentifier>;
```
enthusiastmartin, line 82: you should not need Extended in the pallet implementation.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    pub type Orders<T: Config> = StorageMap<_, Blake2_128Concat, OrderId, Order<T::AccountId, T::AssetId>, OptionQuery>;
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::weight(<T as Config>::WeightInfo::place_order())]
+        #[transactional]
+        pub fn place_order(
+            origin: OriginFor<T>,
+            asset_buy: T::AssetId,
+            asset_sell: T::AssetId,
```
enthusiastmartin, line 168: i'd consider to use in and out instead of sell and buy, so it is consistent with naming across our pallets.

of course, if that would make sense.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+            let order = Order {
+                owner,
+                asset_buy,
+                asset_sell,
+                amount_buy,
+                partially_fillable,
+            };
+
+            Self::validate_place_order(order.clone(), amount_sell)?;
```
enthusiastmartin, line 183: change the validate_place_order to take the order by reference, so there wont be a need to clone it. 

it will make much more sense, since you need to continue using the same order

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+                let current_id = *next_id;
+                *next_id = next_id.checked_add(One::one()).ok_or(Error::<T>::OrderIdOutOfBound)?;
+                Ok(current_id)
+            })?;
+
+            let reserve_id = Self::named_reserve_identifier(order_id);
+            T::Currency::reserve_named(&reserve_id, order.asset_sell, &order.owner, amount_sell)?;
+
+            <Orders<T>>::insert(order_id, order.clone());
+            Self::deposit_event(Event::OrderPlaced {
```
enthusiastmartin, line 195: Switch the order of insert and deposit event, so no need to clone it.

Or use the parameters instead of values from the order object. The order has not changed , that means the input parameters should be correct.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+            Ok(())
+        }
+
+        #[pallet::weight(<T as Config>::WeightInfo::fill_order())]
+        #[transactional]
+        pub fn fill_order(
+            origin: OriginFor<T>,
+            order_id: OrderId,
+            asset_fill: T::AssetId,
+            amount_fill: Balance,
```
enthusiastmartin, line 213: i'd use asset and amount instead of asset_fill, amount_fill - the name of the function is clear enough to understand what asset and amount refers to.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+
+            <Orders<T>>::try_mutate_exists(order_id, |maybe_order| -> DispatchResult {
```
enthusiastmartin, line 217: it should be enough to use try_mutate. The storage if OptionQuery so it should return an option regardless.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+                let order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
+
+                let amount_sell = Self::amount_sell(order_id, order);
+
+                let amount_receive = Self::amount_receive(order, amount_sell, amount_fill)?;
```
enthusiastmartin, line 222: The names of functions `amount_sell` and `amount_receive` are not very clear what they really do here. 

they are the same as the variables, so i am wondering what they supposed to do. I havent yet checked their implemeantion but from just the code reading, it would be nice to understand what they are for.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+                Self::validate_fill_order(order, who.clone(), asset_fill, amount_sell, amount_fill, amount_receive)?;
```
enthusiastmartin, line 224: Same as before - if it takes is be reference, no need for clone.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+                Self::execute_deal(order_id, order, who.clone(), amount_fill, amount_receive)?;
```
enthusiastmartin, line 226: Probably here it would be also good to take the account (who) be reference.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+                let remaining_amount_buy = Self::amount_remaining(order.amount_buy, amount_fill)?;
```
enthusiastmartin, line 228: same comment as for amouint_sell and amount_fill .

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+                if remaining_amount_buy > 0 {
+                    Self::update_storage(order, amount_fill)?;
```
enthusiastmartin, line 231: i am not really a fan of this approach.  becayuse the storage is not really updated as part of the "update_storage" function. 

consider doing something like this:
1. change update_storage to update_order
2. take the order by value
3. within the update_order, update the order 
4. return it back and assign it to maybe_order.

this way, it consumes the order, which is fine, because you no longer need it. and you get back new order. but there wont any moving needed.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        );
+
+        ensure!(
+            T::Currency::can_reserve(order.asset_sell, &order.owner, amount_sell),
+            Error::<T>::InsufficientBalance
+        );
+
+        let min_amount_buy = Self::min_order_size(order.asset_buy)?;
+
+        ensure!(order.amount_buy >= min_amount_buy, Error::<T>::OrderSizeTooSmall);
```
enthusiastmartin, line 295: at first sight, these two lines feel that they do exactlly the same. 

i'd suggest to either:

combine the min buy and ED check into one function call "ensure_min_order_size". 

or

rename it to ensure_ed 

i'd prefer the first.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        let min_amount_sell = Self::min_order_size(order.asset_sell)?;
+
+        ensure!(amount_sell >= min_amount_sell, Error::<T>::OrderSizeTooSmall);
```
enthusiastmartin, line 299: same applies for this one too.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        Ok(())
+    }
+
+    fn validate_fill_order(
+        order: &mut Order<T::AccountId, T::AssetId>,
```
enthusiastmartin, line 305: why mutable is required here ?  i cant figure it out.

regardless, if the order needs to be mutable - the naming of the function is not correct because the naming suggests that it only validates it, but mutable indicates that something is changing too.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        } else {
+            let remaining_amount_buy = Self::amount_remaining(order.amount_buy, amount_fill)?;
+
+            if remaining_amount_buy > 0 {
+                let min_amount_buy = Self::min_order_size(order.asset_buy)?;
+
+                ensure!(
+                    remaining_amount_buy >= min_amount_buy,
+                    Error::<T>::RemainingOrderSizeTooSmall
+                );
```
enthusiastmartin, line 332: as before - ifi you combine this into one - it will be more clear.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    fn named_reserve_identifier(order_id: OrderId) -> [u8; 8] {
+        let prefix = b"otc";
```
enthusiastmartin, line 351: make this as constant somewhere.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        hashed_array
+    }
+
+    fn min_order_size(asset: T::AssetId) -> Result<Balance, Error<T>> {
+        T::ExistentialDeposits::get(&asset)
+            .checked_mul(T::ExistentialDepositMultiplier::get().into())
+            .ok_or(Error::<T>::MathError)
+    }
+
+    fn amount_sell(order_id: OrderId, order: &mut Order<T::AccountId, T::AssetId>) -> Balance {
```
enthusiastmartin, line 368: again - all these function use mutable referecne - it does not seem it is needed.

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    fn amount_remaining(amount_initial: Balance, amount_change: Balance) -> Result<Balance, Error<T>> {
+        amount_initial.checked_sub(amount_change).ok_or(Error::<T>::MathError)
+    }
+
+    #[require_transactional]
+    fn execute_deal(
+        order_id: OrderId,
+        order: &mut Order<T::AccountId, T::AssetId>,
```
enthusiastmartin, line 391: mut?

enthusiastmartin (commented):
Reviewed the main implementation only - i haven't looked at tests or benchmarking yet. 

And only focused mostly from the rust and code quality perspective - i will do logic in another review round once my comments are resolved;-)

otc/src/lib.rs
apopiak: insert should be possible with a `&order`, too

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    fn named_reserve_identifier(order_id: OrderId) -> [u8; 8] {
+        let prefix = b"otc";
+        let mut result = [0; 8];
+        result[0..3].copy_from_slice(prefix);
+        result[3..7].copy_from_slice(&order_id.to_be_bytes());
+
+        let hashed = BlakeTwo256::hash(&result);
+        let mut hashed_array = [0; 8];
+        hashed_array.copy_from_slice(&hashed.as_ref()[..8]);
+        hashed_array
```
apopiak, line 359: what's the benefit of hashing here?

liquidity-mining/Cargo.toml
vgantchev: because impl in hydradx-node depended on another PR which was not merged until this morning

otc/Cargo.toml
vgantchev: removed  balances, math and currencies - can use orml_tokens indeed

traits is used for `Registry`

otc/src/lib.rs
vgantchev: I was thinking about it but _in and _out don't make much sense outside the context of pools imo :)

otc/src/lib.rs
vgantchev: it's by design: try_mutate_exists allows to clean storage if order is fill by setting it to `None`

https://crates.parity.io/frame_support/storage/trait.StorageMap.html#tymethod.try_mutate_exists

otc/src/lib.rs
vgantchev: good suggestion

otc/Cargo.toml
enthusiastmartin: why need of orml-tokens - it should be generic enough so it does not depend on concrete implemenation. 

if used only in tests - then it should be in dev dependencies.

otc/src/lib.rs
enthusiastmartin: yes , kind of agree. but you can also see the in and out as in the protocol, out of the protocol.  Or into the account, out of the account.

otc/src/lib.rs
enthusiastmartin: fair enough

otc/Cargo.toml
vgantchev: this is where it is indeed

otc/src/lib.rs
vgantchev: oki

otc/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+        /// Cannot partially fill an order which is not partially fillable
+        OrderNotPartiallyFillable,
+        /// Order amount_in and amount_out must at all times be greater than the existential deposit
+        /// for the asset multiplied by the ExistentialDepositMultiplier.
+        /// A fill order may not leave behind amounts smaller than this.
+        OrderAmountTooSmall,
+        /// Error with math calculations
+        MathError,
+        /// The caller does not have permission to complete the action
+        NoPermission,
```
enthusiastmartin, line 158: Forbidden ?

otc/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+        ensure!(order.asset_in == asset, Error::<T>::AssetNotInOrder);
+
+        ensure!(order.amount_in >= amount, Error::<T>::CannotFillMoreThanOrdered);
+
+        ensure!(
+            T::Currency::ensure_can_withdraw(asset, who, amount).is_ok(),
+            Error::<T>::InsufficientBalance
+        );
+
+        if !order.partially_fillable {
```
enthusiastmartin, line 340: nitpick: personally, i would check the true first.  meaing `if order.partiall_fillable`.

otc/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+
+                let amount_out = Self::fetch_amount_out(order_id, order);
+
+                let amount_receive = Self::calculate_amount_receive(order, amount_out, amount)?;
+
+                Self::validate_fill_order(order, &who, asset, amount, amount_out, amount_receive)?;
+
+                Self::execute_deal(order_id, order, &who, amount, amount_receive)?;
+
+                let remaining_amount_in = Self::calculate_difference(order.amount_in, amount)?;
```
enthusiastmartin, line 252: you actually calculate the remaining amount twice. first time in validate_fill_order and the second time here. 

i wonder if it could be refactored in some cleaner way.

otc/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+
+    fn calculate_amount_receive(
+        order: &Order<T::AccountId, T::AssetId>,
+        amount_out: Balance,
+        amount_fill: Balance,
+    ) -> Result<Balance, Error<T>> {
+        amount_out
+            .checked_mul(amount_fill)
+            .and_then(|v| v.checked_div(order.amount_in))
+            .ok_or(Error::<T>::MathError)
```
enthusiastmartin, line 388: i believe this multiplication can easily overflow if only u128 is used.

enthusiastmartin (dismissed):
one important issue about the possible multiplication overflow. 

I'd suggest to consider creating some property tests

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    fn validate_fill_order(
+        order: &Order<T::AccountId, T::AssetId>,
+        who: &T::AccountId,
+        asset: T::AssetId,
+        amount: Balance,
+        amount_out: Balance,
+        amount_receive: Balance,
+        remaining_amount_in: Balance,
+    ) -> DispatchResult {
+        ensure!(order.asset_in == asset, Error::<T>::AssetNotInOrder);
```
mrq1911, line 340: this doesn't have to be checked when you don't allow user to set it ;)

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Asset does not exist in registry
+        AssetNotRegistered,
+        /// The asset used to fill the order is different than asset_in of the order
+        AssetNotInOrder,
+        /// When filling and order, the fill amount cannot be greater than the remaining order amount
+        CannotFillMoreThanOrdered,
+        /// Free balance is too low to place the order
+        InsufficientBalance,
```
mrq1911, line 144: this error is redundant as it will be checked and thrown by currencies pallet, therefore there is no need for this error in this pallet at all

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        #[pallet::weight(<T as Config>::WeightInfo::fill_order())]
+        #[transactional]
+        /// Fill an OTC order (partially or completely)
+        ///  
+        /// Parameters:
+        /// - `order_id`: ID of the order
+        /// - `asset`: Asset which is being filled
+        /// - `amount`: amount which is being filled
+        pub fn fill_order(
```
mrq1911, line 233: i would consider splitting it into two functions, one for fully filling the order and the second for partially fill as naively i would assume that fully filling will be cheaper in terms of weight

mrq1911: also, it could only have one simple param, opposed to 2 with partial fill

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        /// - `amount`: Amount which is being filled
+        pub fn cancel_order(origin: OriginFor<T>, order_id: OrderId) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+
+            <Orders<T>>::try_mutate_exists(order_id, |maybe_order| -> DispatchResult {
+                let order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
+
+                ensure!(order.owner == who, Error::<T>::Forbidden);
+
+                let amount_out = Self::fetch_amount_out(order_id, order);
```
mrq1911, line 297: you don't have to fetch this in this case, `unreserve_named` with `Balance::MAX` will unreserve everything reserved. function actually returns what was unserved so you get that value anyway and can use it in `OrderCanceled` event - which would be very interesting information for partially fillable orders

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        }
+
+        #[pallet::weight(<T as Config>::WeightInfo::cancel_order())]
+        #[transactional]
+        /// Cancel an open OTC order
+        ///  
+        /// Parameters:
+        /// - `order_id`: ID of the order
+        /// - `asset`: Asset which is being filled
+        /// - `amount`: Amount which is being filled
```
mrq1911, line 288:
```suggestion
```

otc/src/types.rs
```diff
@@ -0,0 +1,30 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use codec::{Decode, Encode, MaxEncodedLen};
+use scale_info::TypeInfo;
+
+pub type Balance = u128;
+pub type NamedReserveIdentifier = [u8; 8];
+pub type OrderId = u32;
```
mrq1911, line 21: i guess that is a disadvantage of using named reserves for accounting to have only 2^32 them :)

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        OrderFilled {
+            order_id: OrderId,
+            who: T::AccountId,
+            amount: Balance,
+        },
+        /// An Order has been partially filled
+        OrderPartiallyFilled {
+            order_id: OrderId,
+            who: T::AccountId,
+            amount: Balance,
```
mrq1911, line 121: amount_in

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+            amount_receive: Balance,
```
mrq1911, line 122: amount_out

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(crate) fn deposit_event)]
+    pub enum Event<T: Config> {
+        /// An Order has been cancelled
+        OrderCancelled { order_id: OrderId },
+        /// An Order has been completely filled
+        OrderFilled {
+            order_id: OrderId,
+            who: T::AccountId,
+            amount: Balance,
```
mrq1911, line 115: amount_in, amount_out should be added for consistency as well

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        result
+    }
+
+    fn fetch_amount_out(order_id: OrderId, order: &Order<T::AccountId, T::AssetId>) -> Balance {
+        let reserve_id = Self::named_reserve_identifier(order_id);
+        T::Currency::reserved_balance_named(&reserve_id, order.asset_out, &order.owner)
+    }
+
+    fn calculate_amount_receive(
```
mrq1911, line 388: calculate_filled_amount_out

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        who: &T::AccountId,
+        asset: T::AssetId,
+        amount: Balance,
+        amount_out: Balance,
+        amount_receive: Balance,
+        remaining_amount_in: Balance,
+    ) -> DispatchResult {
+        ensure!(order.asset_in == asset, Error::<T>::AssetNotInOrder);
+
+        ensure!(order.amount_in >= amount, Error::<T>::CannotFillMoreThanOrdered);
```
mrq1911, line 342: i would allow this to make sure if somebody fills a portion before you when you want to fully fill, you get at least something with your fill instead of failed tx and delayed full fill for maker, so you would fill the order not with exact amount_in but `up to` amount_in

mrq1911 (changes requested):

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        #[pallet::constant]
+        type ExistentialDepositMultiplier: Get<u8>;
+
+        /// Weight information for the extrinsics.
+        type WeightInfo: WeightInfo;
+    }
+
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(crate) fn deposit_event)]
+    pub enum Event<T: Config> {
```
mrq1911, line 108: prefix Order on all events is potentially redundant

otc/src/lib.rs
vgantchev: hmm not sure - this adds quite some complexity (from code, testing but also UX pov), while the benefits of addressing this edge case are questionable

I would revisit it in a future iteration if that's ok with you?

jak-pan: Would maybe create separate callable function for this? (partial fill)

otc/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+            origin: OriginFor<T>,
+            order_id: OrderId,
+            asset: T::AssetId,
+            amount: Balance,
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+
+            <Orders<T>>::try_mutate_exists(order_id, |maybe_order| -> DispatchResult {
+                let order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
+
```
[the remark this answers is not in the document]
vgantchev: that's everywhere in warehouse, uses tabs for identation :-(

otc/src/lib.rs
vgantchev: The weight should not be very different - in the one case you mutate a value, in the other you mutate the whole record to None

I am not a big fan of splitting it because partial_fill would still need to be able to handle full fill, can't trust frontend that it will call correct function also it can be exploited

otc/src/lib.rs
vgantchev: @jak-pan even if we have a separate extrinsic for partial_fill, it still needs to be able to handle a complete fill imo

Otherwise attackers can dispatch a complete fill via partial_fill, which will lead to unfreeze of the balance which is used as storage bond, without cleaning the record :(

vgantchev: or we can add an extra validation to partial_fill that it must leave at least 5xED, if you think that splitting in 2 is rly worth it

otc/src/lib.rs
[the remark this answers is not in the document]
mrq1911: i was reffering to the vertical spacing - e.g. extra lines inserted into code

otc/src/lib.rs
mrq1911: i thought it returns only what it really unreserved, nevermind then

otc/src/tests/mock.rs
```diff
@@ -0,0 +1,238 @@
...
+
+thread_local! {
+    pub static DUMMYTHREADLOCAL: RefCell<u128> = RefCell::new(100);
+}
+
+pub fn expect_events(e: Vec<Event>) {
+    test_utils::expect_events::<Event, Test>(e);
+}
+
+pub fn named_reserve_identifier(order_id: OrderId) -> [u8; 8] {
```
mrq1911, line 232: i would use the same implementation as is in the pallet, no need to copypsate. also i would note that this yield unreadable strings for UI
```
String.fromCharCode(...[111, 116, 99, 0, 0, 0, 1, 0])
'otc\x00\x00\x00\x01\x00'
```

mrq1911: also currently runtime lets you have only 50 reserves per account, so with this it could be pretty easy to exceed this limit. i am leaning towards not abusing this identification for accounting and do the internal accounting in pallet instead, as it will solve 3 issues:
- increase number of global orders to u128::MAX from u32::MAX
- unreadable reserve id
- reserve id generation on ui
- limited orders per account to 50

vgantchev: Here storage is very cheap and non-problematic, a good reason to no waste named reserves on this

DCA is up to your consideration, i already left a comment on the PR

otc/src/lib.rs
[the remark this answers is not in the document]
vgantchev: i usually try to logically separate code with empty lines for better readability, but in the case that you commented it was excessive indeed :)

otc/src/lib.rs
```diff
@@ -300,19 +292,14 @@ pub mod pallet {
...
-
             <Orders<T>>::try_mutate_exists(order_id, |maybe_order| -> DispatchResult {
                 let order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
-
                 ensure!(order.owner == who, Error::<T>::Forbidden);
 
-                let reserve_id = Self::named_reserve_identifier(order_id);
-                T::Currency::unreserve_named(&reserve_id, order.asset_out, &order.owner, Balance::MAX);
-
+                T::Currency::unreserve(order.asset_out, &order.owner, order.amount_out);
```
mrq1911, line 299: i would still used named reserve, but with just constant identifier to separate what user is reserving in the otc vs dca, etc

vgantchev: ok i will change back to named with const but this was not very explicit in your previous comments :)

otc/src/tests/invariants.rs
```diff
@@ -0,0 +1,67 @@
...
+    min..max
+}
+
+prop_compose! {
+    fn get_asset_amounts()
+        (
+            amount_in in asset_amount(6 * ONE, 100 * ONE),
+        )
+        (
+            amount_in in Just(amount_in),
```
enthusiastmartin, line 32: it is not possible to have amount out < amount in at all ? 

if that;'s the case, then it is ok. otherwise, you'd want to have such scenarios covered too.

otc/src/tests/invariants.rs
```diff
@@ -0,0 +1,67 @@
...
+            OTC::place_order(
+                Origin::signed(ALICE),
+                DAI,
+                HDX,
+                initial_amount_in,
+                initial_amount_out,
+                true
+            ).unwrap();
+
+            let initial_price = initial_amount_out / initial_amount_in;
```
enthusiastmartin, line 57: i'd use some fixed type here.

mrq1911 (changes requested):

otc/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            amount_in: Balance,
+            amount_out: Balance,
+            partially_fillable: bool,
+        },
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Asset does not exist in registry
+        AssetNotRegistered,
```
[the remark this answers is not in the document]
vgantchev: @mrq1911 this is only true for asset_out which is reserved - otherwise nothing prevents you from creating an order for a non-existent asset_in, with all issues down the stream (UI).

Please LMK if I should still remove

otc/src/benchmarks.rs
```diff
@@ -0,0 +1,121 @@
...
+        assert_ok!(
+            crate::Pallet::<T>::place_order(RawOrigin::Signed(owner.clone()).into(), dai.into(), hdx.into(), 20 * ONE, 100 * ONE, true)
+        );
+  }:  _(RawOrigin::Signed(owner.clone()), 0u32)
+    verify {
+        assert_eq!(T::Currency::reserved_balance_named(&NAMED_RESERVE_ID, hdx.into(), &owner), 0);
+    }
+}
+
+fn prepare<T: Config>() -> Result<(u32, u32), DispatchError>
```
dmoka, line 82: I would name it as "create_assets_in_registry"

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+            let order = Order {
+                owner,
+                asset_in,
+                asset_out,
+                amount_in,
+                amount_out,
+                partially_fillable,
+            };
+
+            // Validations
```
dmoka, line 209: instead of having a comment, I would just wrap all the 4 validations to a method "validate_new_order" method, with 2 arguments: order and amount_out

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+            ensure!(T::AssetRegistry::exists(order.asset_in), Error::<T>::AssetNotRegistered);
+            ensure!(
+                T::Currency::can_reserve(order.asset_out, &order.owner, amount_out),
+                Error::<T>::InsufficientBalance
+            );
+            Self::validate_min_order_amount(order.asset_in, order.amount_in)?;
+            Self::validate_min_order_amount(order.asset_out, amount_out)?;
+
+            let order_id = <NextOrderId<T>>::try_mutate(|next_id| -> result::Result<OrderId, DispatchError> {
```
dmoka, line 218: the way it is implemented is on a different abstraction level than what it functionally does.

I would extract it to a method called "get_next_order_id".

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+                let id = *next_id;
```
dmoka, line 219: Do we start at id 0? Should not we start at id 1?

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+                Ok(())
+            })
+        }
+
+        #[pallet::weight(<T as Config>::WeightInfo::fill_order())]
+        #[transactional]
+        /// Fill an OTC order (completely)
+        ///  
+        /// Parameters:
+        /// - `order_id`: ID of the order
```
dmoka, line 280: I guess you miss the origin in the API doc.

dmoka: It is also true for other extrinsics

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub type NextOrderId<T: Config> = StorageValue<_, OrderId, ValueQuery>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn orders)]
+    pub type Orders<T: Config> = StorageMap<_, Blake2_128Concat, OrderId, Order<T::AccountId, T::AssetId>, OptionQuery>;
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::weight(<T as Config>::WeightInfo::place_order())]
+        #[transactional]
```
dmoka, line 182: AFAIK, this transactional annotation is not needed anymore. 

also true for other extrinsics

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+        /// Create a new OTC order
+        ///  
+        /// Parameters:
+        /// - `asset_in`: Asset which is being bought
+        /// - `asset_out`: Asset which is being sold
+        /// - `amount_in`: Amount that the order is seeking to buy
+        /// - `amount_out`: Amount that the order is selling
+        /// - `partially_fillable`: Flag indicating whether users can fill the order partially
```
dmoka, line 190: As far as I see in other pallets, we usually also document that emitted events like

```
		/// Emits `<EventName>` event when <status>.
```

If you do, please do it in other extrinsics as well.

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+            .checked_mul(T::ExistentialDepositMultiplier::get().into())
+            .ok_or(Error::<T>::MathError)?;
+
+        ensure!(amount >= min_amount, Error::<T>::OrderAmountTooSmall);
+
+        Ok(())
+    }
+
+    fn calculate_difference(amount_initial: Balance, amount_change: Balance) -> Result<Balance, Error<T>> {
+        amount_initial.checked_sub(amount_change).ok_or(Error::<T>::MathError)
```
dmoka, line 346: Maybe `ArithmeticError::Underflow` would be better? If so, you could use Underflow/Overflow in other places too, so we can get rid of this customer math error.

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    #[require_transactional]
+    fn execute_deal(
```
dmoka, line 350: `execute_order` ?

otc/src/tests/place_order.rs
```diff
@@ -0,0 +1,147 @@
...
+#[test]
+fn place_order_should_throw_error_when_amount_out_is_too_low() {
+    ExtBuilder::default().build().execute_with(|| {
+        // Act
+        assert_noop!(
+            OTC::place_order(Origin::signed(ALICE), DAI, HDX, 20 * ONE, 4 * ONE, true),
+            Error::<Test>::OrderAmountTooSmall
+        );
+    });
+}
```
dmoka, line 147: Awesome and clean tests! Also went through the code with mutation testing, and all mutations are killed! Great job!

otc/src/lib.rs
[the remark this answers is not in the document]
mrq1911: we can keep it then

otc/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+        /// Fill an OTC order (completely)
+        ///  
+        /// Parameters:
+        /// - `order_id`: ID of the order
+        pub fn fill_order(origin: OriginFor<T>, order_id: OrderId) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+            let order = <Orders<T>>::get(order_id).ok_or(Error::<T>::OrderNotFound)?;
+
+            Self::execute_deal(&order, &who, order.amount_in, order.amount_out)?;
+            <Orders<T>>::remove(order_id);
```
mrq1911, line 286: i am not sure that this actually removes the key from storage as well, not according to @nohaapav

otc/src/lib.rs
vgantchev: it was in a separate method and was moved to in-function per @mrq1911 request

otc/src/lib.rs
vgantchev: as discussed during call - not possible, probably bc remapping brings O(n)

otc/src/lib.rs
vgantchev: yup, and we also haven't documented origin in other features (consistency)

mrq1911 (approved):

PR #166, opened 2023-02-13
feat: pallet duster

martinfridrich:
This PR is adding pallet-duster from Basilisk repo so this pallet can be reused for Hydra-dx.


enthusiastmartin (approved):
as it is the original duster, let's merge.

however, i'd suggest complete or at least partial refactor.

PR #169, opened 2023-02-21
feat: dca

dmoka:
Extracting some helper and common logic to this repo, for the DCA functionality


traits/src/pools.rs
```diff
@@ -19,3 +21,31 @@ pub trait DustRemovalAccountWhitelist<AccountId> {
...
+    fn buy(
+        origin: Origin,
+        asset_in: AssetId,
+        asset_out: AssetId,
+        amount: Balance,
+        max_sell_amount: Balance,
+    ) -> DispatchResult;
+}
+
+pub trait PriceProvider<AssetId> {
```
enthusiastmartin, line 44: why we need two different traits for price provider which it seems would do exactly the same ?

relaychain-info/src/lib.rs
```diff
@@ -46,6 +46,10 @@ pub mod pallet {
     #[pallet::error]
     pub enum Error<T> {}
 
+    #[pallet::storage]
+    #[pallet::getter(fn parent_hash)]
+    pub(super) type ParentHash<T> = StorageValue<_, Hash, ValueQuery>;
```
enthusiastmartin, line 51: could you add some description behind the motivation to store this value ?

vgantchev:
should we have whole dca pallet in warehouse? allowing to eventually reuse for Snek?

route-executor/src/lib.rs
```diff
@@ -381,6 +365,31 @@ impl<T: Config> Pallet<T> {
     }
 }
 
+impl<T: Config> TradeAmountsCalculator<T::AssetId, T::Balance> for Pallet<T> {
```
mrq1911, line 368: is this still used with the new version with router?

traits/src/pools.rs
mrq1911: @dmoka ?

mrq1911 (changes requested):

dmoka:
> should we have whole dca pallet in warehouse? allowing to eventually reuse for Snek?

as we will have the mono repo soon, this is not necessary anymore

route-executor/src/lib.rs
dmoka: yes, it was added lately, because we need to determine the exact amount_in for buy orders.

Instead of duplicating the logic in hydra, I just exposed this function by using the trait.

mrq1911:
move it to the hydradx dca PR when we have monorepo

PR #171, opened 2023-02-22
feat: fungible trait implementation for currencies 

enthusiastmartin:
Adds implementations of fungible traits for currencies adapter.

### DIFF
--- a/currencies/src/fungibles.rs
+++ b/currencies/src/fungibles.rs
@@ -5,7 +5,7 @@ use frame_support::traits::tokens::{fungible, fungibles, DepositConsequence, Wit
 use orml_traits::MultiCurrency;
 use sp_runtime::traits::Get;
 use sp_runtime::DispatchError;
-use std::marker::PhantomData;
+use sp_std::marker::PhantomData;
 
 pub struct FungibleCurrencies<T>(PhantomData<T>);
 
--- a/currencies/src/fungibles.rs
+++ b/currencies/src/fungibles.rs
@@ -1,0 +1,146 @@
+use crate::module::{BalanceOf, CurrencyIdOf};
+use crate::{Config, Pallet};
+use frame_support::dispatch::DispatchResult;
+use frame_support::traits::tokens::{fungible, fungibles, DepositConsequence, WithdrawConsequence};
+use orml_traits::MultiCurrency;
+use sp_runtime::traits::Get;
+use sp_runtime::DispatchError;
+use std::marker::PhantomData;
+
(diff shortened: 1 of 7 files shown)


currencies/src/fungibles.rs
```diff
@@ -0,0 +1,146 @@
...
+    <T::NativeCurrency as fungible::Inspect<T::AccountId>>::Balance: Into<BalanceOf<T>> + From<BalanceOf<T>>,
+    WithdrawConsequence<BalanceOf<T>>:
+        From<WithdrawConsequence<<T::NativeCurrency as fungible::Inspect<T::AccountId>>::Balance>>,
+    Result<Self::Balance, DispatchError>:
+        From<Result<<T::NativeCurrency as fungible::Inspect<T::AccountId>>::Balance, DispatchError>>,
+    Result<Self::Balance, DispatchError>:
+        From<Result<<T::MultiCurrency as fungibles::Inspect<T::AccountId>>::Balance, DispatchError>>,
+{
+    fn mint_into(asset: Self::AssetId, who: &T::AccountId, amount: Self::Balance) -> DispatchResult {
+        <Pallet<T>>::deposit(asset, who, amount)
```
Roznovjak, line 91: why not `mint_into` methods?

enthusiastmartin: it is actually the same. but it can be changed to use mint into, so it is clear.

Roznovjak (approved):

currencies/src/tests_fungibles.rs
```diff
@@ -0,0 +1,82 @@
...
+                true
+            ));
+            assert_eq!(FungibleCurrencies::<Runtime>::balance(NATIVE_CURRENCY_ID, &ALICE), 90);
+            assert_eq!(FungibleCurrencies::<Runtime>::balance(NATIVE_CURRENCY_ID, &BOB), 10);
+
+            assert_ok!(FungibleCurrencies::<Runtime>::transfer(
+                X_TOKEN_ID, &BOB, &ALICE, 10, true
+            ));
+            assert_eq!(FungibleCurrencies::<Runtime>::balance(X_TOKEN_ID, &BOB), 90);
+            assert_eq!(FungibleCurrencies::<Runtime>::balance(X_TOKEN_ID, &ALICE), 10);
```
enthusiastmartin, line 80: can you actually use directly pallet balances and orml tokens storage to verify that it was correctly updated after transfer? 

similar to what it is done in the other tests.

enthusiastmartin: also i would add test with keep alive false, and ensur that it works too.

Issue #179, opened 2023-03-09
Suggestions to simplify state change calculations


yanliu18:
https://github.com/galacticcouncil/warehouse/blob/b67141958f31fc3af0894a22b5534a26e67f6334/liquidity-mining/src/lib.rs#L1639-L1650

In the above code, there are a few reasons to simplify,
1. The above code contains duplicated computation which could reduce the precision of results and waste computation power (large number division is considered expensive, though it might be negligible in the case of single execution). In detail, in the above implementation, 
let $V_y$ = `yield_farm.total_valued_shares`, 
     $r_y$ = `yield_farm_reward`, 
     $m_y$ = `yield_farm.multiplier`, 
    $rpz_g$ = `global_farm.accumulated_rpz`, 
    similarly, for `yield_farm.accumulated_rpz` and others, we have
$$r_y = V_y * m_y * (rpz_g - rpz_y)$$
$$\Delta rpvs_y = r_y / V_y = m_y * (rpz_g - rpz_y)$$
$$rpvs_y' = rpvs_y +  \Delta rpvs_y $$
Thus, there is no need to calculate the intermediate variable `stake_in_global_farm` which is $V_y * m_y$.  

2. To flatten the call structure, the functions `calculate_global_farm_shares` and `calculate_rewards_from_pot` can be removed.
3. Keeping the storage writes in one place would give better readability- another reason to remove `calculate_rewards_from_pot`.

In the end, the suggested sudo code is as follows:
```rust
        let (delta_rpvs, yield_farm_rewards) = math::calculate_yield_farm_rewards(
            yield_farm.accumulated_rpz,
            global_farm.accumulated_rpz,
            yield_farm.multiplier,
            yield_farm.total_value_shares,
        ) ... // error handling

        yield_farm.accumulated_rpvs = yield_farm
            .accumulated_rpvs.
            .checked_add(delta_rpvs)
            .ok_or(ArithmeticError::Overflow)?;

        yield_farm.accumulated_rpz = global_farm.accumulated_rpz;

       // move other yield farm updates here.

        global_farm.accumulated_paid_rewards = global_farm
            .accumulated_paid_rewards
            .checked_add(yield_farm_rewards)
            .ok_or(ArithmeticError::Overflow)?;

        global_farm.pending_rewards = global_farm
            .pending_rewards
            .checked_sub(yield_farm_rewards)
            .ok_or(ArithmeticError::Overflow)?;

```
The above code is not tested, please modify properly when adding to production code.

PR #185, opened 2023-03-16
feat: lm calculation simplification and refactor

martinfridrich:
* #179 
* https://github.com/galacticcouncil/HydraDX-math/issues/92

I didn't move `periods_since_last_update` calculation to math repo to preserve `InconsistentState` error and to not  complicate `MathError` from math repo.


liquidity-mining/src/tests/tests.rs
```diff
@@ -863,389 +863,6 @@ fn sync_global_farm_should_not_update_farm_when_farm_was_already_updated_in_this
     }
 }
 
-#[test]
-fn claim_from_global_farm_should_work() {
```
apopiak, line 867: this test is no longer useful?

apopiak (approved):
looks fine to me

liquidity-mining/src/tests/tests.rs
martinfridrich: function tested by this test(`calculate_rewards_from_pot()`) was removed and its functionality was moved to math repo and inlined into the `sync_global_farm()` function so this test is not necessary anymore.

PR #186, opened 2023-03-16
chore: add new oracle period

apopiak:
This PR adds a new oracle period with a = 0.2 (which corresponds to a 9 block oracle when calculating with `smoothing_from_period`).

Based on #182


enthusiastmartin (approved):
the benchmarking changes are part of another PR.

PR #189, opened 2023-03-21
fix: liquidity mining valued shares >= min_deposit

### LINKED ISSUE #184: `global_farm.add_stake` could add 0 stakes to the global farm while the `valued_shares` are non-zero
In the functions `update_yield_farm_multiplier`, `resume_yield_farm` and `do_deposit_lp_shares`, when adding the `valued_shares` to the global farm, it is not assured that `math::calculate_global_farm_shares(yield_farm.total_valued_shares, multiplier)` returns a non-zero value when `yield_farm.total_valued_shares` is non-zero and `multiplier` is ensured to be no less than `MIN_YIELD_FARM_MULTIPLIER`. 

For the functions `update_yield_farm_multiplier`, `resume_yield_farm`, shares could be withdrawn beforehand to bring the `total_valued_shares` below the `global_farm.min_deposit`.

As for the function  `do_deposit_lp_shares`, the following checking
https://github.com/galacticcouncil/warehouse/blob/b67141958f31fc3af0894a22b5534a26e67f6334/liquidity-mining/src/lib.rs#L1357
won't guarantee the `valued_shares`  calculated at 
https://github.com/galacticcouncil/warehouse/blob/b67141958f31fc3af0894a22b5534a26e67f6334/liquidity-mining/src/lib.rs#L1374
to be no less than `global_farm.min_deposit` since the spot price of the deposit token could be any value.

Thus, it is recommended to add the check `ensure!(valued_shares >= global_farm.min_deposit, Error::<T, I>::InvalidDepositAmount);` before depositing the valued shares to the global farm.

This checking could be wrapped with the call to `math::calculate_global_farm_shares` into an internal function `add_stake_to_global_farm` to encapsulate the stake-adding logic as a whole.

> Though it might be acceptable to ignore small deposits to the global farm, i.e., the result of `math::calculate_global_farm_shares(valued_shares, yield_farm.multiplier)` is rounded to 0, it is not feasible in the current implementation when the `valued_shares` are subject to price manipulation risk.

martinfridrich:
This PR fixes #184

### DIFF
--- a/liquidity-mining/Cargo.toml
+++ b/liquidity-mining/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "3.2.0"
+version = "4.0.0"
 description = "Liquidity mining"
 authors = ["GalacticCouncil"]
 edition = "2021"
--- a/liquidity-mining/src/lib.rs
+++ b/liquidity-mining/src/lib.rs
@@ -293,8 +293,8 @@ pub mod pallet {
         /// Account creation from id failed.
         ErrorGetAccountId,
 
-        /// Value of deposited shares amount in reward currency can't be 0.
-        ZeroValuedShares,
+        /// Value of deposited shares amount in reward currency is bellow min. limit.
+        InvalidValuedShares,
 
         /// `reward_currency` is not registered in asset registry.
         RewardCurrencyNotRegistered,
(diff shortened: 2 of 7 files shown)


liquidity-mining/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "3.2.0"
+version = "4.0.0"
```
enthusiastmartin, line 3: it does not feel like major change

liquidity-mining/src/lib.rs
```diff
@@ -1376,7 +1376,12 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
                         deposit.shares,
                     )?;
 
-                    ensure!(valued_shares.gt(&Balance::zero()), Error::<T, I>::ZeroValuedShares);
+                    //NOTE: this is not InconsistentStateError. This may happen if shares are
```
enthusiastmartin, line 1379: i feel like the comment is not really necessary

enthusiastmartin (approved):

liquidity-mining/Cargo.toml
martinfridrich: It's breaking change in the API, error changed.

PR #192, opened 2023-03-28
refactor: otc minor improvemenst + invariants test with various precisions

enthusiastmartin:


otc/src/lib.rs
```diff
@@ -318,15 +315,17 @@ pub mod pallet {
...
         pub fn cancel_order(origin: OriginFor<T>, order_id: OrderId) -> DispatchResult {
             let who = ensure_signed(origin)?;
             <Orders<T>>::try_mutate_exists(order_id, |maybe_order| -> DispatchResult {
                 let order = maybe_order.as_mut().ok_or(Error::<T>::OrderNotFound)?;
 
-                // Validations
                 ensure!(order.owner == who, Error::<T>::Forbidden);
 
-                T::Currency::unreserve_named(&NAMED_RESERVE_ID, order.asset_out, &order.owner, order.amount_out);
+                let unreserved =
```
vgantchev, line 326: this variable name is misleading - the fn will return a non-zero value only if the requested amount is higher than reserved amount

maybe call it sth along the lines of `amount_failed_to_unreserve`

otc/src/lib.rs
```diff
@@ -359,7 +354,8 @@ impl<T: Config> Pallet<T> {
         amount_out: Balance,
     ) -> DispatchResult {
         T::Currency::transfer(order.asset_in, who, &order.owner, amount_in)?;
-        T::Currency::unreserve_named(&NAMED_RESERVE_ID, order.asset_out, &order.owner, amount_out);
+        let unreserved = T::Currency::unreserve_named(&NAMED_RESERVE_ID, order.asset_out, &order.owner, amount_out);
```
vgantchev, line 357: better variable name, same as above

otc/src/lib.rs
enthusiastmartin: renamed to "remaining_to_unreserve". hope that is slightly better.

vgantchev (approved):
LGTM

PR #195, opened 2023-04-03
feat: arbitrary price oracle

dmoka:


traits/src/oracle.rs
```diff
@@ -19,6 +19,10 @@ impl<AssetId, Price> NativePriceOracle<AssetId, Price> for () {
     }
 }
 
+pub trait PriceOracle<AssetId, Price> {
```
apopiak, line 22: docs?

apopiak (approved):

traits/src/oracle.rs
enthusiastmartin, line 22: i'd suggest to do Price as associated type here.

dmoka: I can do that. But why would it be better?

PR #198, opened 2023-04-18
chore : update to latest math for omnipool's remove liquid

enthusiastmartin:


Roznovjak (approved):
Bump crate versions

mrq1911:
@enthusiastmartin how are these changes related to the omnipool fix, these pallets doesn;t use omnipool math for anything as far as i know

enthusiastmartin:
> @enthusiastmartin how are these changes related to the omnipool fix, these pallets doesn;t use omnipool math for anything as far as i know?

You have to use the same math version because now, there is a type EmaPrice used in runtime configuration. 

So if you don't use the same version, it starts complaining that a trait is not implemented for that, as it sees it as different type.

Agree that this looks weird. But is necessary.

Monorepo ftw.

PR #200, opened 2023-04-19
Feat: add xcm rate limit to registry

dmoka:


asset-registry/src/lib.rs
```diff
@@ -192,6 +192,7 @@ pub mod pallet {
                 asset_type: AssetType::Token,
                 existential_deposit: self.native_existential_deposit,
                 locked: false,
+                xcm_rate_limit: None,
```
dmoka, line 195: Please consider adding rate limit to the genesis config.

PR #201, opened 2023-05-03
Feat:LM price_adjustment fallback when oracle is not available

martinfridrich:
This PR implements fallback for liquidity-mining `price_adjustment` when oracle is not  available.
We are saving last value of `price_adjustment` from oracle in the `GlobalFarm` and if oracle is not available this last saved value should be used instead of error.

### DIFF
--- a/liquidity-mining/src/tests/lm_with_oracle.rs
+++ b/liquidity-mining/src/tests/lm_with_oracle.rs
@@ -162,3 +162,137 @@ fn non_full_farm_should_pay_rewards_with_half_speed_when_price_adjustmnet_is_fro
         });
     });
 }
+
+#[test]
+fn oracles_price_adjustment_should_be_used_and_saved_when_oracle_is_available() {
+    new_test_ext().execute_with(|| {
+        let _ = with_transaction(|| {
+            const GLOBAL_FARM: GlobalFarmId = 1;
+            const YIELD_FARM_A: YieldFarmId = 2;
+
+            const ALICE_DEPOSIT: DepositId = 1;
+
+            const TOTAL_REWARDS: u128 = 200_000 * ONE;
+
+            //Arrange
+            set_block_number(100);
+            assert_ok!(LiquidityMining3::create_global_farm(
+                TOTAL_REWARDS,
+                20,
+                10,
+                BSX,
+                BSX,
+                GC,
+                Perquintill::from_float(0.5),
+                1_000,
+                FixedU128::from_float(5.5_f64), //default price_adjustment, won't be used.
+            ));
+
+            assert_ok!(LiquidityMining3::create_yield_farm(
+                GC,
+                GLOBAL_FARM,
+                FixedU128::from(2_u128),
+                None,
+                BSX_TKN1_AMM,
+                vec![BSX, TKN1],
+            ));
+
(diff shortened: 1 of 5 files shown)


liquidity-mining/src/tests/lm_with_oracle.rs
```diff
@@ -162,3 +162,137 @@ fn non_full_farm_should_pay_rewards_with_half_speed_when_price_adjustmnet_is_fro
...
+            assert_ok!(LiquidityMining3::create_global_farm(
+                TOTAL_REWARDS,
+                20,
+                1,
+                BSX,
+                BSX,
+                GC,
+                Perquintill::from_float(0.5),
+                1_000,
+                FixedU128::from(2_u128), //default price_adjustment, won't be used.
```
Roznovjak, line 251: *will be used

liquidity-mining/src/tests/lm_with_oracle.rs
```diff
@@ -162,3 +162,137 @@ fn non_full_farm_should_pay_rewards_with_half_speed_when_price_adjustmnet_is_fro
...
+            set_block_number(999_666_334);
+            //NOTE: Oracle will fail because global_farm.updated_at == 999_666_333 when oracle is
+            //called.
+            let (_, _, claimed_amount, unclaimable) =
+                LiquidityMining3::claim_rewards(ALICE, ALICE_DEPOSIT, YIELD_FARM_A, false).unwrap();
+
+            //Assert
+            assert_eq!(unclaimable, 0);
+            assert_eq!(claimed_amount, 5_000 * ONE);
+            //NOTE: global-farm's price_adjustment should be updated
```
Roznovjak, line 289: * last saved value is used

Roznovjak (approved):

apopiak:
The idea is to set a manual value in case an oracle is not available?
Cause it's unlikely that you suddenly won't get any value from the oracle (except if it was removed from storage somehow).

liquidity-mining/src/tests/lm_with_oracle.rs
martinfridrich: Comment is correct, `FixedU128::from(2_u128)` id "default price_adjustment" and it will not be used in the test. `price_adjustment` from oracle is used in the test.

liquidity-mining/src/tests/lm_with_oracle.rs
martinfridrich: I'm not sure what you mean. Comment is for assert bellow and note that `price_adjustment` should be updated.

martinfridrich:
> The idea is to set a manual value in case an oracle is not available?
> Cause it's unlikely that you suddenly won't get any value from the oracle (except if it was removed from storage somehow).

Yes, @jak-pan wanted to have a fallback if for some reason oracle won't be able to provide price.

jak-pan:
> The idea is to set a manual value in case an oracle is not available?
> Cause it's unlikely that you suddenly won't get any value from the oracle (except if it was removed from storage somehow).

Yes. In any case it's not available we want to be sure you are able to withdraw liquidity. If the oracles don't work for any reason we would lock out users in the positions.

jak-pan (approved):

PR #204, opened 2023-05-09
feat: rework fee payment

enthusiastmartin:
This PR refactors the way that fee is withdrawn for the set_currency call. 

set_currency no longer deducts fees as part of the extrinsics itself. Instead, OnChargeTransaction checks if processed tx is set_currency. If so, the currency parameter is used a fee currency.

Other changes:
- WithdrawFees struct removed completely
- MultiDataProvider trait removed as it only introduced complexity.
- several pallet's config parameters removed as they are no longer necessary

### DIFF
--- a/traits/src/pools.rs
+++ b/traits/src/pools.rs
@@ -5,9 +5,10 @@ pub trait SpotPriceProvider<AssetId> {
 
     /// Return spot price for given asset pair.
     ///
-    /// Returns price of the `asset_b` denominated in `asset_a` ( `asset_a`/`asset_b` ).
+    /// Returns price of the `asset_b` denominated in `asset_a` ( `asset_a / asset_b` ).
+    /// Example: `spot_price(DAI, LRNA) == 25` (you get 25 DAI for each LRNA).
     ///
-    /// Returns None if such pair does not exist.
+    /// Returns `None` if such pair does not exist.
     fn spot_price(asset_a: AssetId, asset_b: AssetId) -> Option<Self::Price>;
 }
 


martinfridrich (approved):
LGTM

traits/src/pools.rs
```diff
@@ -3,9 +3,11 @@ pub trait SpotPriceProvider<AssetId> {
...
     fn pair_exists(asset_a: AssetId, asset_b: AssetId) -> bool;
 
-    /// Return spot price for given asset pair
+    /// Return spot price for given asset pair.
     ///
-    /// Returns None if such pair does not exist
+    /// Returns price of the `asset_b` denominated in `asset_a` ( `asset_a`/`asset_b` ).
+    ///
+    /// Returns None if such pair does not exist.
     fn spot_price(asset_a: AssetId, asset_b: AssetId) -> Option<Self::Price>;
```
apopiak, line 11: note for other reviewers: we clarified this in a way that is the inverse from the previous semantics (it used to be implemented as b/a in Hydra. So we might want to check what Basilisk expects here.

traits/src/pools.rs
```diff
@@ -3,9 +3,11 @@ pub trait SpotPriceProvider<AssetId> {
 
     fn pair_exists(asset_a: AssetId, asset_b: AssetId) -> bool;
 
-    /// Return spot price for given asset pair
+    /// Return spot price for given asset pair.
     ///
-    /// Returns None if such pair does not exist
+    /// Returns price of the `asset_b` denominated in `asset_a` ( `asset_a`/`asset_b` ).
+    ///
+    /// Returns None if such pair does not exist.
```
apopiak, line 10:
```suggestion
    /// Returns price of the `asset_b` denominated in `asset_a` ( `asset_a / asset_b` ).
    /// Example: `spot_price(DAI, LRNA) == 25` (you get 25 DAI for each LRNA).
    ///
    /// Returns `None` if such pair does not exist.
```

apopiak (approved):
LGTM, but I was part of writing it, so would be good to get another review

transaction-multi-payment/src/tests.rs
```diff
@@ -1073,9 +837,8 @@ fn fee_withdrawal_cannot_kill_account_when_paid_in_native() {
     const CHARLIE: AccountId = 5;
 
     ExtBuilder::default()
-        .account_native_balance(CHARLIE, 30)
+        .account_native_balance(CHARLIE, 3)
```
Roznovjak, line 840: I think that with this change we are testing a wrong error. We get `InsufficientBalance` instead of `KeepAlive`.

PR #206, opened 2023-05-18
fix: multicurrency payment fee rouding

apopiak:
This PR ensures that a non-zero fee is charged even if the calculated fee in the native currency is small and the non-native fee currency is expensive (low price value).

### DIFF
--- a/transaction-multi-payment/src/tests.rs
+++ b/transaction-multi-payment/src/tests.rs
@@ -215,6 +215,40 @@ fn fee_payment_in_non_native_currency() {
 }
 
 #[test]
+fn fee_payment_in_expensive_non_native_currency() {
+    const CHARLIE: AccountId = 5;
+
+    ExtBuilder::default()
+        .base_weight(5)
+        .account_tokens(CHARLIE, HIGH_VALUE_CURRENCY, 10_000)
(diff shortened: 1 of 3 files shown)


transaction-multi-payment/src/tests.rs
```diff
@@ -214,6 +214,40 @@ fn fee_payment_in_non_native_currency() {
         });
 }
 
+#[test]
+fn fee_payment_in_expensive_non_native_currency() {
```
Roznovjak, line 218:
```suggestion
fn fee_payment_in_expensive_non_native_currency_should_be_non_zero() {
```

transaction-multi-payment/src/tests.rs
```diff
@@ -214,6 +214,40 @@ fn fee_payment_in_non_native_currency() {
...
+    const CHARLIE: AccountId = 5;
```
Roznovjak, line 219: Use `Bob`. He's defined in the mock file, and he _ain't got a penny!_ :)

Roznovjak (approved):

mrq1911 (approved):

mrq1911:
LGTM

PR #211, opened 2023-05-30
Fix: use same math version everywhere

apopiak:


martinfridrich (approved):
please bump patch version of LM pallet otherwise LGTM

